PR #48, opened 2020-04-28
LAWS-823-Alter-cpu-memory-and-timeout

donmasters:
- Restore Load Balancer timout value
- Increase Memory and CPU values

This is required as tests are showing that cpu is fully utilized on
startup causing the Load Balancer checks to fail. Will evaluate and
request load testing to finalise.


donmasters:
This approach is no longer required

PR #84, opened 2020-06-04
Feature/add preconidtion check controller

praveenmoj:
## What

 Add new validation URI for linking.

https://dsdmoj.atlassian.net/browse/CACP-377

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/exception/RestControllerAdviser.java
```diff
@@ -1,36 +1,51 @@
...
+     *
+     * @param ex
+     * @return
+     */
+    @ExceptionHandler({MAATCourtDataException.class})
+    public ResponseEntity<Object> handleRuntimeException(RuntimeException ex, WebRequest request) {
+
+        return handleExceptionInternal(ex,
+                ErrorDTO.builder()
+                        .message(ex.getMessage())
```
atharmajeed, line 32: do we need to set a code value too?

praveenmoj: correct, we have to set code values once integrated upstream, i had a discussion with Dan from VCD and richard. The messages we send might not be exactly what business wants to see. So when they start integrating the validation endpoints they will come with codes for each fails cases and we send them the code along with a brief description. The VCD will be having a business specific message which also can support english/welsh. Then we have to set the error codes as agreed, we might have a discussion around that time and change accordingly

PR #85, opened 2020-06-08
Refactoring Changes

msahilan07:
## What

Code Improvement 
Adding Tests

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/model/LaaTransactionLogging.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/model/LaaTransactionLogging.java
@@ -13,9 +13,9 @@ import java.util.UUID;
 @Value
 public class LaaTransactionLogging {
 
-    private UUID laaTransactionId;
-    private Integer maatId;
-    private String caseUrn;
+    UUID laaTransactionId;
+    Integer maatId;
+    String caseUrn;
 
 
 }
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/controller/UnLinkController.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/controller/UnLinkController.java
@@ -2,33 +2,32 @@ package gov.uk.courtdata.unlink.controller;
 
 import gov.uk.courtdata.model.Unlink;
 import gov.uk.courtdata.unlink.validator.UnLinkValidationProcessor;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
 @RestController
 @RequestMapping("/unlink")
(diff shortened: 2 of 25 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/validator/UserIdValidator.java
```diff
@@ -16,7 +16,7 @@
     @Override
     public Optional<Void> validate(String userId) throws ValidationException {
 
-        if (!userId.isEmpty() && userId.length()>0) {
+        if (userId != null) {
```
atharmajeed, line 19: Don't we need to check if the user-id is an empty string or not?

msahilan07: Good Point. Fixed it now!

maat-court-data-api/src/test/java/gov/uk/courtdata/unlink/validator/UserIdValidatorTest.java
```diff
@@ -0,0 +1,28 @@
...
+    private UserIdValidator userIdValidator;
+
+
+
+    @Test
+    public void testUserIdValidator_whenUserIsNullThrowsException() {
+
+        thrown.expect(ValidationException.class);
+        thrown.expectMessage("User id is missing.");
+        userIdValidator.validate(null);
```
atharmajeed, line 26: we might need to add another test for empty string.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/LaaTransactionLogging.java
```diff
@@ -13,9 +13,9 @@
 @Value
 public class LaaTransactionLogging {
 
-    private UUID laaTransactionId;
-    private Integer maatId;
-    private String caseUrn;
+    UUID laaTransactionId;
```
atharmajeed, line 16: Just being pedantic here, we should restrict the access fields and they should only be accessible within a class, the default value is package-private unless there is a reason I'm not aware of. I'm fine either way.

maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/controller/UnLinkController.java
```diff
@@ -2,33 +2,32 @@
...
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
 @RestController
 @RequestMapping("/unlink")
 @Slf4j
+@RequiredArgsConstructor
```
atharmajeed, line 14: This is good.

maat-court-data-api/src/test/java/gov/uk/courtdata/unlink/validator/UserIdValidatorTest.java
msahilan07: Added, also added to reason ID validator.

PR #102, opened 2020-06-26
Feature/392 CP hearing message retry

atharmajeed:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-392)

When processing a CP Hearing notification, if the MAAT Record is locked then system put the message back the hearing queue with a delay. The message will be processed later by the consumer. 

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/hearing/service/HearingResultedPublisher.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Gson gson = new Gson();
+        String hearingResultedJSON = gson.toJson(hearingResulted);
+
+        AmazonSQS amazonSQS = amazonSQSConfig.awsSqsClient();
+
+        SendMessageRequest request = new SendMessageRequest()
+                .withQueueUrl(amazonSQS.getQueueUrl(sqsQueueName).getQueueUrl())
+                .withMessageBody(hearingResultedJSON)
+                .withDelaySeconds(60)
```
atharmajeed, line 50: I'll change this to 900 seconds (this is the max we can delay a message.)

maat-court-data-api/src/main/java/gov/uk/courtdata/util/LaaTransactionLoggingBuilder.java
```diff
@@ -14,10 +15,15 @@ private LaaTransactionLoggingBuilder() {
     public static LaaTransactionLogging get(CaseDetails caseDetails) {
         Gson gson = new Gson();
         return get(gson.toJson(caseDetails));
+    }
 
+    public static LaaTransactionLogging get(HearingResulted hearingResulted) {
+        Gson gson = new Gson();
```
msahilan07, line 21: Need a test case for this change.

PR #113, opened 2020-06-30
Feature/392 hearing message retry

atharmajeed:
## What

When processing a CP Hearing notification, if the MAAT Record is locked then system put the message back the hearing queue with a delay. The message will be processed later by the consumer.

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/hearing/crowncourt/CrownTestController.java
```diff
@@ -0,0 +1,21 @@
+package gov.uk.courtdata.hearing.crowncourt;
```
msahilan07, line 1: This should not be in the repo. Please delete this. :)

maat-court-data-api/src/test/java/gov/uk/courtdata/integrationTest/MockServicesConfig.java
```diff
@@ -34,4 +35,7 @@
     @MockBean(name = "cdaOAuth2WebClient")
     private WebClient cdaOAuth2WebClient;
 
+    @MockBean
+    private HearingResultedPublisher hearingResultedPublisher;
```
msahilan07, line 39: We dont Mock the classes that we we worte. We mock third party Beans (eg: AmazonSQSConfig which actually needs the  mocking within HearingResultedPublisher )

atharmajeed: Well spotted and totally agreed.

PR #126, opened 2020-07-09
Added unit test cases for LaaStatusUpdate.

atharmajeed:
## What

Missing unit test cases for laasstatusupdate. For builder classes. 

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/model/laastatus/LaaStatusUpdate.java
```diff
@@ -2,14 +2,13 @@
...
-import lombok.AllArgsConstructor;
-import lombok.Builder;
-import lombok.NoArgsConstructor;
+import lombok.*;
 
 
 @Builder
 @NoArgsConstructor
 @AllArgsConstructor
+@Data
```
praveenmoj, line 11: The object can be build using Builder and data annotation will allow for modification after creation which we dont want when generate json requests. Could we remove the @Data annotation, please?

atharmajeed: Changed to @Value.

PR #127, opened 2020-07-13
Notification audit

praveenmoj:
## What

-  Extend jms message logging to all listeners
- Schedule house keeping cron job to purge message log entries as configured for the no of days.

## Checklist


maat-court-data-api/src/test/java/gov/uk/courtdata/service/QueueMessageLogServiceTest.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @Test
+    public void whenUnLinkMessage_CheckLogEntryCreated() {
+
+        final Integer maatId = 1000;
+
+        queueMessageLogService.createLog(QueueMessageType.UNLINK, newQueueMessage(maatId));
+
+        verify(queueMessageLogRepository, times(1)).save(queueMessageCaptor.capture());
```
atharmajeed, line 65: Just a suggestion, the default value for times(1) is one so we don't have to explicitly add this param (times).

praveenmoj: Thanks, I can take it off.

PR #132, opened 2020-07-24
Feature/derive imprisonment flag

praveenmoj:
## What

- Added tests for imprisonment
- Tidy up result code matching.

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/enums/CCTrialOutcome.java
```diff
@@ -14,15 +15,15 @@
...
 
-    private String outcome;
+    private String value;
 
     public static boolean isConvicted(String outcome) {
 
-        Optional.ofNullable(outcome).orElseThrow(
-                () -> new IllegalArgumentException("Crown Court outcome can't be empty."));
+        String out = Optional.ofNullable(outcome).orElseThrow(
+                () -> new ValidationException("Crown Court outcome can't be empty."));
```
atharmajeed, line 23: 👍

PR #149, opened 2020-08-06
PR for Open API, swagger and header value

atharmajeed:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-424)

Added configuration for open API and swagger 3.
Made header fields as optional. 

image

## Checklist


maat-court-data-api/src/main/resources/application.yaml
```diff
@@ -21,12 +21,19 @@ spring:
             clientId: ${CDA_OAUTH_CLIENT_ID}
             clientSecret: ${CDA_OAUTH_CLIENT_SECRET}
             authorizationGrantType: client_credentials
+
 springdoc:
+  packagesToScan: gov.uk.courtdata
+  show-actuator: true
   api-docs:
     path: /maat-api
+    enabled: true
```
atharmajeed, line 30: This parameter is to enable/disable the API pages. We might want to disable for prod. In this case, I think we should move this as an external variable. What do you guys think?

PR #183, opened 2020-09-18
Sonarqube - remove scm provider and to use artifact from s3

praveenmoj:
https://dsdmoj.atlassian.net/browse/CACP-509

sonarqube - remove scm provider property and allow to use artefact from s3.

## Checklist


donmasters (approved):
LGTM

PR #191, opened 2020-09-25
Update wq core create and processed to timestamp

praveenmoj:
On events in MLRA created and processed time doesn't include timestamp. 

https://dsdmoj.atlassian.net/browse/CACP-520

 Fix to add time to created and processed date on events.

## Checklist


atharmajeed (approved):
👍🏼

PR #317, opened 2021-07-13
Refactored codepipeline to have the approvals after the changeset and before execution

alanrana7:
## What

Amends the manual approval stages in the code pipeline so that the change set is built before the manual approval which allows the changes to be double checked before executing the change set to different environments. (Not needed in development).

Jira - https://dsdmoj.atlassian.net/browse/LAWS-1788

## Checklist


liamraeAL (approved):
Looks good

atharmajeed (approved):

PR #328, opened 2021-08-17
removed the sqs listener. There are error code 400 type errors for sq…

atharmajeed:
…s not found.

## What

## Checklist


laurentqro (approved):

msahilan07 (changes requested):
This is the wrong listener.  We shouldn't be removing this. I guess we should be removing LaaStatusJobListener. Let's have a call.

nyenugulamoj (approved):

nyenugulamoj (approved):

msahilan07 (approved):

PR #389, opened 2021-12-07
Adding schema for the initial means assessment processing

msahilan07:
## What

 https://dsdmoj.atlassian.net/browse/LASB-775

Schema Creation for the Initial Means Assessment processing. 

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/resources/schema/apiCreateInitialMeansAssessment.json
+++ b/maat-court-data-api/src/main/resources/schema/apiCreateInitialMeansAssessment.json
@@ -112,7 +112,7 @@
   },
   "additionalProperties": false,
   "required": [
-    "maatId",
+    "repId",
     "laaTransactionId",
     "cmuId",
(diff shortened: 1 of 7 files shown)


atharmajeed (approved):

Mhowell494 (commented):
I can't see where the MaatID is required to be stored, or is that the same as the RepOrderID?

The assessmentID is generated by a sequence in the database, so I don't think we need to be sending this for the initial assessment.

What is the correlationID?

The reviewType schema isn’t included. Looking at the version in the means assessment pull request, it to contains a required field: description, is that needed here? I can’t see where it needs to be stored.

employmentCode should probably be changed to employmentStatus.

The totalAggregatedIncome and adjustedIncomeValue are missing, is the idea to calculate these from the assessment details fields?

The initResult and initResultReason fields are missing.

Can we include the schemas for each of the the individual objects?

msahilan07:
>

PR #401, opened 2022-01-05
CRU operations for Passport assessments with delete disabled in API

rajkrk:
Create/Update/Get operations for Passport Assessments, with Delete not exposed via API

## What

https://dsdmoj.atlassian.net/browse/LASB-743

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/controller/PassportAssessmentController.java
```diff
@@ -0,0 +1,94 @@
...
+        gson = gsonBuilder
+                .setPrettyPrinting()
+                .serializeNulls()
+                .registerTypeAdapter(LocalDateTime.class, new DateUtil.LocalDateTimeAdapter())
+                .create();
+    }
+
+    @GetMapping(value = "/{passportAssessmentId}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Retrieve a passport assessment record")
+    @ApiResponses(value = {
```
Mhowell494, line 51: This can be removed, the @ApiResponse annotation is repeatable now. We probably need to go back and change this in the existing controller too.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/controller/PassportAssessmentController.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @PutMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Update a passport assessment record")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE)),
+            @ApiResponse(responseCode = "400", description = "Bad Request.", content = @Content(mediaType = "application/json", schema = @Schema(implementation = ErrorDTO.class))),
+            @ApiResponse(responseCode = "500", description = "Server Error.", content = @Content(mediaType = "application/json", schema = @Schema(implementation = ErrorDTO.class)))
+    })
+    public ResponseEntity<Object> updateAssessment(
+            @Parameter(description = "Passport assessment data", content = @Content(mediaType = "application/json",
```
Mhowell494, line 72: There is a mix of MediaType.APPLICATION_JSON_VALUE and application/json, it might be better if we stuck with the prior when specifying all media types.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/controller/PassportAssessmentController.java
```diff
@@ -0,0 +1,94 @@
...
+        passportAssessmentValidationProcessor.validate(passportAssessmentId);
+        PassportAssessmentDTO passportAssessment = passportAssessmentService.find(passportAssessmentId);
+        String passportAssessmentJson = gson.toJson(passportAssessment, PassportAssessmentDTO.class);
+        return ResponseEntity.ok(passportAssessmentJson);
+    }
+
+    @PutMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Update a passport assessment record")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE)),
```
Mhowell494, line 67: Can we include the response schema for a successful request?

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/controller/PassportAssessmentController.java
```diff
@@ -0,0 +1,94 @@
...
+@Tag(name = "Passport Assessments", description = "Rest API for passport assessments")
+public class PassportAssessmentController {
+
+    private Gson gson;
+    private final GsonBuilder gsonBuilder;
+    private final PassportAssessmentService passportAssessmentService;
+    private final PassportAssessmentValidationProcessor passportAssessmentValidationProcessor;
+
+    @PostConstruct
+    private void initialize() {
```
Mhowell494, line 41: I wonder if we're also doing this in the other controller, is there a way to centralise this logic to avoid duplication?

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/impl/FinancialAssessmentImpl.java
```diff
@@ -73,9 +73,8 @@ public FinancialAssessmentEntity create(FinancialAssessmentDTO financialAssessme
     }
 
     public void setOldAssessmentReplaced(FinancialAssessmentDTO financialAssessment) {
-        Integer repID = financialAssessment.getRepId();
-        financialAssessmentRepository.updateOldAssessments(repID);
-        passportAssessmentRepository.updateOldPassportAssessments(repID);
-        hardshipReviewRepository.updateOldHardshipReviews(repID, financialAssessment.getId());
+        financialAssessmentRepository.updatePreviousFinancialAssessmentsAsReplaced(financialAssessment.getRepId(), financialAssessment.getId());
```
Mhowell494, line 76: This shouldn't match on the id. This is the relevant snippet from the existing PL/SQL:

`update financial_assessments f`
`          set replaced = 'Y'`
     `   where F.REP_ID = p_application_object.rep_id`
         ` and f.id != p_application_object.current_assessment_object.fin_assessment_object.fin_assessment_id;`

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
```diff
@@ -0,0 +1,55 @@
...
+    private String dwpResult;
+    private String passportNote;
+    private String between16And17;
+    private String under18HeardInYouthCourt;
+    private String under18HeardInMagsCourt;
+    private LocalDateTime lastSignOnDate;
+    private String esa;
+    private String pastStatus;
+    @Builder.Default
+    private String replaced = "N";
```
Mhowell494, line 42: This is already handled by the entity, it doesn't need to be part of the schema.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
```diff
@@ -0,0 +1,55 @@
...
+    private LocalDateTime passportEvidenceDueDate;
+    private LocalDateTime allPassportEvidenceReceivedDate;
+    private Integer passportUpliftPercentage;
+    private LocalDateTime passportUpliftApplyDate;
+    private LocalDateTime passportUpliftRemoveDate;
```
Mhowell494, line 47: The following fields all relate to passport assessment evidence, which isn't handled by the current implementation. I think for now these could probably be removed:

passportEvidenceDueDate
allPassportEvidenceReceivedDate
passportUpliftPercentage
passportUpliftApplyDate
passportUpliftRemoveDate
passportEvidenceNotes

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/CreatePassportAssessment.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.time.LocalDateTime;
+
+@Data
+@SuperBuilder
+@AllArgsConstructor
+@NoArgsConstructor
+@EqualsAndHashCode(callSuper = true)
+public class CreatePassportAssessment extends PassportAssessment {
+    private Integer financialAssessmentId;
```
Mhowell494, line 17: I can't see why this field is required? If associated financial assessments need updating then we can do so using the RepID.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
```diff
@@ -0,0 +1,61 @@
...
+    public void delete(Integer passportAssessmentId) {
+        passportAssessmentImpl.delete(passportAssessmentId);
+    }
+
+    @Transactional(rollbackFor = MAATCourtDataException.class)
+    public PassportAssessmentDTO create(CreatePassportAssessment createPassportAssessment) {
+        log.info("Create Passport Assessment - Transaction Processing - Start");
+        PassportAssessmentDTO passportAssessmentDTO =
+                passportAssessmentMapper.createPassportAssessmentToPassportAssessmentDTO(createPassportAssessment);
+        log.info("Creating new passport assessment record");
```
Mhowell494, line 48: Before creating a new assessment, we should first check for existing incomplete passport/hardship assessments for the rep order and raise an error if any exist.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
```diff
@@ -0,0 +1,55 @@
...
+    private String passportEvidenceNotes;
+    private LocalDateTime firstPassportReminderDate;
+    private LocalDateTime secondPassportReminderDate;
+    private String valid;
+    private LocalDateTime dateCompleted;
```
Mhowell494, line 52: This field doesn't need to be passed. The value inserted into the database should be based on the past_status field, if it's 'COMPLETE' the current date is used.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
```diff
@@ -0,0 +1,61 @@
...
+        return buildPassportAssessmentDTO(passportAssessmentEntity);
+    }
+
+    @Transactional(rollbackFor = MAATCourtDataException.class)
+    public PassportAssessmentDTO update(UpdatePassportAssessment updatePassportAssessment) {
+        log.info("Update Passport Assessment - Transaction Processing - Start");
+        PassportAssessmentDTO passportAssessmentDTO =
+                passportAssessmentMapper.updatePassportAssessmentToPassportAssessmentDTO(updatePassportAssessment);
+        log.info("Updating existing passport assessment record");
+        PassportAssessmentEntity passportAssessmentEntity = passportAssessmentImpl.update(passportAssessmentDTO);
```
Mhowell494, line 34: There needs to be a check here to determine whether the passport assessments status is marked as complete, as then it should not be modified.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/impl/PassportAssessmentImpl.java
```diff
@@ -0,0 +1,75 @@
...
+        passportAssessmentRepository.deleteById(passportAssessmentId);
+    }
+
+    public PassportAssessmentEntity create(PassportAssessmentDTO passportAssessmentDTO) {
+        PassportAssessmentEntity passportAssessmentEntity = assessmentMapper.passportAssessmentDtoToPassportAssessmentEntity(passportAssessmentDTO);
+        return passportAssessmentRepository.save(passportAssessmentEntity);
+    }
+
+    public void setOldPassportAssessmentAsReplaced(PassportAssessmentDTO passportAssessment) {
+        passportAssessmentRepository.updatePreviousPassportAssessmentsAsReplaced(passportAssessment.getRepId(), passportAssessment.getId());
```
Mhowell494, line 71: This shouldn't match on the ID, the assessments we want to mark as replaced are those with a matching RepID but with a DIFFERENT id.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidator.java
```diff
@@ -0,0 +1,33 @@
...
+@Slf4j
+@AllArgsConstructor
+@Component
+public class CreatePassportAssessmentValidator implements IValidator<Void, CreatePassportAssessment> {
+
+    @Override
+    public Optional<Void> validate(CreatePassportAssessment createPassportAssessment) {
+
+        if (isBlank(createPassportAssessment.getNworCode())) {
+            throw new ValidationException("New work reason code is required");
```
Mhowell494, line 23: This is being done in the validation processor already.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidator.java
```diff
@@ -0,0 +1,33 @@
...
+        } else if (isBlank(createPassportAssessment.getUserCreated())) {
+            throw new ValidationException("Username is required");
+        } else if (isBlank(createPassportAssessment.getPastStatus())) {
+            throw new ValidationException("Past Status is required");
```
Mhowell494, line 27: This looks like it's required for both create and update requests, therefore I would consider moving it to the validation processor.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/validator/UpdatePassportAssessmentValidatorTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Mock
+    private PassportAssessmentIdValidator passportAssessmentIdValidator;
+
+    @Before
+    public void setUp() {
+        when(passportAssessmentIdValidator.validate(any())).thenReturn(Optional.empty());
+    }
+
+    @Test
+    public void testUpdateAssessmentValidator_whenUserModifiedIsBlank_thenThrowsException() {
```
Mhowell494, line 35: Should we also check blanks for the main validation processor?

Mhowell494 (commented):
There seems to be quite a lot of missing business logic when creating/updating, particularly around assessment evidence.

Is the intention for this to be implemented in another ticket? If so, can it be created now?

There are a few other minor issues (see comments), but otherwise it looks great.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/CreatePassportAssessment.java
rajkrk: This is the reason :  
update TOGDATA.HARDSHIP_REVIEWS hr
         set
            HR.REPLACED = 'Y'
         where
            HR.FIAS_ID  = p_application_object.current_assessment_object.fin_assessment_object.**fin_assessment_id**;

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
rajkrk: This is the parent class for create/update objects which only have fields specific to them.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
rajkrk: These will be needed once the passport evidence APIs are added in. Since we do not know how the business logic would be implemented in means assessment, it might be better to leave these in

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
rajkrk: The above logic would be in means assessment service so this has to be passed in

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
rajkrk: I think these would go to means assessment service as it is business logic, as part of validations

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/controller/PassportAssessmentController.java
rajkrk: This has been removed as there is no actual need for gson (it was only being used to convert to String but now we use objects and let Spring take care of converstion

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/CreatePassportAssessment.java
Mhowell494: Well spotted. I wonder if there's a way to retrieve the latest financial assessment without needing it? Best to leave it for now though I think.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
Mhowell494: I think it makes more sense to include that logic in the meat-ap. To me, as the application that manages the database interaction, it should also be responsible for insuring the integrity of the data being inserted/updated.

maat-court-data-api/src/main/java/gov/uk/courtdata/model/assessment/PassportAssessment.java
Mhowell494: Yes, but the only time replaced is ever set to 'Y' is when a new one is created with a matching RepID, which is already being handled. The entity then sets replaced to 'N' by default, so I don't think we don't need to expose the field.

`@Builder.Default`
`@Column(name = "REPLACED")`
`private String replaced = "N";`

maat-court-data-api/src/main/java/gov/uk/courtdata/repository/PassportAssessmentRepository.java
```diff
@@ -15,6 +15,6 @@
     void updateAllPreviousPassportAssessmentsAsReplaced(@Param("repId") Integer repId);
 
     @Modifying
-    @Query(value = "UPDATE PassportAssessmentEntity pa set pa.replaced = 'Y' WHERE pa.id = :id AND pa.repId = :repId")
+    @Query(value = "UPDATE PassportAssessmentEntity pa set pa.replaced = 'Y' WHERE pa.id <> :id AND pa.repId = :repId")
```
Mhowell494, line 18: Native queries should be used to ensure consistency with existing repositories.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidator.java
```diff
@@ -19,13 +19,9 @@
...
 
-        if (isBlank(createPassportAssessment.getNworCode())) {
-            throw new ValidationException("New work reason code is required");
-        } else if (isBlank(createPassportAssessment.getUserCreated())) {
+        if (isBlank(createPassportAssessment.getUserCreated())) {
             throw new ValidationException("Username is required");
-        } else if (isBlank(createPassportAssessment.getPastStatus())) {
-            throw new ValidationException("Past Status is required");
-        } if (createPassportAssessment.getFinancialAssessmentId() == null) {
+        } else if (createPassportAssessment.getFinancialAssessmentId() == null) {
```
Mhowell494, line 24: There should be a check to reject create requests with a passport assessment ID.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
```diff
@@ -31,6 +32,13 @@ public PassportAssessmentDTO update(UpdatePassportAssessment updatePassportAsses
         PassportAssessmentDTO passportAssessmentDTO =
                 passportAssessmentMapper.updatePassportAssessmentToPassportAssessmentDTO(updatePassportAssessment);
         log.info("Updating existing passport assessment record");
+        PassportAssessmentEntity existingPassportAssessmentEntity = passportAssessmentImpl.find(passportAssessmentDTO.getId());
+        if(existingPassportAssessmentEntity == null) {
+            throw new ValidationException(String.format("Passport assessment with id %s not found !",passportAssessmentDTO.getId()));
+        }
+        if(existingPassportAssessmentEntity.getPastStatus().equals(STATUS_COMPLETE)) {
+            throw new ValidationException("User cannot modify a completed assessment");
+        }
```
Mhowell494, line 41: There is a missing check here that verifies the assessment has not been modified by another user.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
```diff
@@ -31,6 +32,13 @@ public PassportAssessmentDTO update(UpdatePassportAssessment updatePassportAsses
         PassportAssessmentDTO passportAssessmentDTO =
                 passportAssessmentMapper.updatePassportAssessmentToPassportAssessmentDTO(updatePassportAssessment);
         log.info("Updating existing passport assessment record");
+        PassportAssessmentEntity existingPassportAssessmentEntity = passportAssessmentImpl.find(passportAssessmentDTO.getId());
```
Mhowell494, line 35: This check is redundant as it is already being done in one of the validator classes.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/controller/PassportAssessmentControllerTest.java
```diff
@@ -121,19 +120,13 @@ public void givenIncorrectParameters_whenUpdatePassportAssessmentIsInvoked_thenE
...
-    @Test
-    public void givenCorrectParameters_whenDeletePassportAssessmentIsInvoked_thenPassportAssessmentIsDeleted() throws Exception {
-        when(passportAssessmentValidationProcessor.validate(any(Integer.class))).thenReturn(Optional.empty());
-        doNothing().when(passportAssessmentService).delete(MOCK_ASSESSMENT_ID);
-        mvc.perform(MockMvcRequestBuilders.delete(endpointUrl + "/" + MOCK_ASSESSMENT_ID))
-                .andExpect(status().isOk());
-    }
 
     @Test
     public void givenIncorrectParameters_whenDeletePassportAssessmentIsInvoked_thenErrorIsThrown() throws Exception {
```
Mhowell494, line 125: This function should probably be renamed or removed. Does it make sense to have a test when there is no controller method?

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/service/PassportAssessmentServiceTest.java
```diff
@@ -90,6 +92,42 @@ public void whenUpdateIsInvoked_thenAssessmentIsUpdated() {
...
+        when(passportAssessmentService.buildPassportAssessmentDTO(any())).thenReturn(
+                PassportAssessmentDTO.builder().id(1000).build()
+        );
+
+        ValidationException validationException = Assert.assertThrows(ValidationException.class,
+                () -> passportAssessmentService.update(passportAssessment));
+        assertThat(validationException.getMessage()).isEqualTo("Passport assessment with id 20000 not found !");
+    }
+
+    public void whenUpdateIsInvokedOnCompletedPassportAssessment_thenValidationExceptionIsThrown() {
```
Mhowell494, line 111: The @Test annotation is missing here.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/service/PassportAssessmentServiceTest.java
```diff
@@ -90,6 +92,42 @@ public void whenUpdateIsInvoked_thenAssessmentIsUpdated() {
         assertThat(returnedAssessment.getId()).isEqualTo(1000);
     }
 
+    @Test
+    public void whenUpdateIsInvokedWithInvalidId_thenValidationExceptionIsThrown() {
+        PassportAssessmentDTO passportAssessmentDTO = TestModelDataBuilder.getPassportAssessmentDTO();
+        passportAssessmentDTO.setId(20000);
+        UpdatePassportAssessment passportAssessment = TestModelDataBuilder.getUpdatePassportAssessment();
+        PassportAssessmentEntity existingPassportAssessmentEntity = TestEntityDataBuilder.getPassportAssessmentEntity();
```
Mhowell494, line 100: This variable isn't being used.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidatorTest.java
```diff
@@ -37,20 +37,6 @@ public void testCreatePassportAssessmentValidator_whenNworCodeIsFilled_thenValid
         assertThat(result).isEqualTo(Optional.empty());
     }
 
-    @Test
```
Mhowell494: The remaining test that checks the Nwor code should also be removed.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidatorTest.java
```diff
@@ -37,20 +37,6 @@ public void testCreatePassportAssessmentValidator_whenNworCodeIsFilled_thenValid
...
-        assertThat(validationException.getMessage()).isEqualTo("New work reason code is required");
-    }
-
-    @Test
-    public void testCreatePassportAssessmentValidator_whenNworCodeIsNull_thenThrowsException() {
-        ValidationException validationException = Assert.assertThrows(ValidationException.class,
-                () -> createAssessmentValidator.validate(getAssessmentWithNworCode(null)));
-        assertThat(validationException.getMessage()).isEqualTo("New work reason code is required");
-    }
-
```
Mhowell494: There are no tests that check the financial assessment ID validation.

Mhowell494 (changes requested):
I've added several comments that should be addressed. I also identified the following issues:

- The setOldPassportAssessmentAsReplaced method should not call the existing updateOldHardshipReviews. Instead a new method should be created that doesn't utilise the REP_ID as it is not required here.

- There is missing validation around passport assessment creation. This logic can be found in the following stored procedures: assessments.check_outstanding_assessment, hardship.check_outstanding_hardship and check_passported_details.

- The update_app_date_completed stored procedure that is called when creating as assessment has not been migrated.

- The handle_passport_result stored procedure has not been migrated. This is called after updating/creating a passport assessment.

maat-court-data-api/src/test/java/gov/uk/courtdata/assessment/controller/PassportAssessmentControllerTest.java
rajkrk: I was trying to assert that calls to delete will fail but I have removed this now

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/validator/CreatePassportAssessmentValidator.java
rajkrk: There is no id field in the CreatePassportAssessment object, so cannot be checked

rajkrk:
> I've added several comments that should be addressed. I also identified the following issues:
> 
> * The setOldPassportAssessmentAsReplaced method should not call the existing updateOldHardshipReviews. Instead a new method should be created that doesn't utilise the REP_ID as it is not required here.
> * There is missing validation around passport assessment creation. This logic can be found in the following stored procedures: assessments.check_outstanding_assessment, hardship.check_outstanding_hardship and check_passported_details.
> * The update_app_date_completed stored procedure that is called when creating as assessment has not been migrated.
> * The handle_passport_result stored procedure has not been migrated. This is called after updating/creating a passport assessment.

This has now been clarified as being out of scope for this particular story and other tickets have been raised for migrating business logic

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/service/PassportAssessmentService.java
rajkrk: This seems to be part of PL/SQL so will have to be addressed as part of migrating that code. This check is not present in FinancialAssessments and IOJ Appeals services either

Mhowell494 (approved):

PR #408, opened 2022-01-19
Lasb 801

Mhowell494:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-801)

Implemented a new reservation status endpoint along with the associated business logic.

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/constants/CourtDataConstants.java
```diff
@@ -32,4 +34,7 @@ private CourtDataConstants() {
     public static final String RESULT_CODE_1000 = "1000";
     public static final Integer APPLICATION_ASN_SEQ_INITIAL_VALUE = 1000;
 
+    public static final String RESERVATION_RECORD_NAME = "REP_ORDER";
+    public static final List<String> RESERVATION_SPECIAL_USERNAMES = List.of("HUB", "MLA");
```
atharmajeed, line 38: this could be an enum to avoid typos etc.

atharmajeed, line 38: this could be an enum to avoid typos etc.

atharmajeed (approved):
👍🏼

PR #433, opened 2022-02-10
Feature/lasb 867 migrate check outstanding assessments

rajkrk:
## What

https://dsdmoj.atlassian.net/browse/LASB-867

Migrated the check_outstanding_assessments function and added an API to invoke the migrated function using the rep id as input from means assessment service. Added unit tests

## Checklist

- [x ] You should have checked that the commit messages say why the change was made.

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
@@ -1,14 +1,18 @@
 package gov.uk.courtdata.dto;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+
 @Data
 @Builder
 @AllArgsConstructor
 @NoArgsConstructor
+@JsonInclude(NON_NULL)
 public class OutstandingAssessmentResultDTO {
     @Builder.Default
     boolean outstandingAssessments = false;
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
@@ -1,12 +1,15 @@
 package gov.uk.courtdata.dto;
 
+
 import com.fasterxml.jackson.annotation.JsonInclude;
+
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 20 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
```diff
@@ -0,0 +1,16 @@
...
+
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class OutstandingAssessmentResultDTO {
+    @Builder.Default
+    boolean outstandingAssessments = false;
+    String message;
+}
```
AafAnsari, line 16: along with the default `outstandingAssessments ` value as false, I think it would be nice to have a default message as well in the response in place of null. 
`{
    "outstandingAssessments": false,
    "message": null
}`

maat-court-data-api/src/main/java/gov/uk/courtdata/repository/FinancialAssessmentRepository.java
```diff
@@ -21,4 +21,8 @@
     @Modifying
     @Query(value = "UPDATE FinancialAssessmentEntity fa set fa.replaced = 'Y' WHERE fa.id <> :id AND fa.repId = :repId")
     void updatePreviousFinancialAssessmentsAsReplaced(@Param("repId") Integer repId, @Param("id") Integer id);
+
+    @Query(value = "SELECT count(*) FROM FINANCIAL_ASSESSMENTS fa WHERE fa.REP_ID = :repId AND fa.REPLACED = 'N' AND (fa.VALID IS NULL OR fa.VALID <> 'N') AND (fa.FASS_FULL_STATUS = 'IN PROGRESS' OR fa.FASS_INIT_STATUS  = 'IN PROGRESS')", nativeQuery = true)
+    Long findOutstandingFinancialAssessments(@Param("repId") Integer repId);
+
```
AafAnsari, line 27: is the native query used here because the sql is read-only ?

AafAnsari (changes requested):

maat-court-data-api/src/main/java/gov/uk/courtdata/dto/OutstandingAssessmentResultDTO.java
rajkrk: I have added a Jackson annotation to exclude empty fields as the message field won't be used by the consumer of this API when there are no outstanding assessments

maat-court-data-api/src/main/java/gov/uk/courtdata/repository/FinancialAssessmentRepository.java
rajkrk: A previous review comment from Matt on PR 401 asked to use native queries to ensure code is consistent

AafAnsari (approved):

PR #517, opened 2022-06-13
LASB-1095 - Add integration tests for authorization controller

mtac50:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-1095)

- [x] Added integration tests for the Autorization controller endpoints.

## Checklist


AafAnsari (changes requested):
do you think it's worth adding the assert for status code 400 and 500 for negative test cases ?

mtac50:
> 

Good point, the way I've set up the tests doesn't actually trigger the RestController that handles the error cases. I'll use the MockMVC to call the method via its endpoint rather than calling it directly.

AafAnsari (approved):

PR #532, opened 2022-06-22
LASB-998

Mhowell494:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-998)

- Enabled AWS x-ray tracing for all endpoints, listeners and scheduled tasks
- Updated CloudFormation templates to run x-ray daemon in sidecar


maat-court-data-api/src/main/java/gov/uk/courtdata/repository/FinancialAssessmentDetailsRepository.java
```diff
@@ -2,9 +2,11 @@
 
 import gov.uk.courtdata.entity.FinancialAssessmentDetailEntity;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
 
 import java.util.List;
 
+@Repository
```
mtac50, line 9: I'm guessing this is just for consistency with the other classes? Technically its not required when the class extends `JpaRepository`.

mtac50 (approved):

alanrana7 (approved):
Cloudformation template looks good to me 👍

maat-court-data-api/src/main/java/gov/uk/courtdata/repository/FinancialAssessmentDetailsRepository.java
Mhowell494: Yep, there is also a pointcut in the x-ray spring aop library that checks for the annotation before it enables tracing on repositories.

PR #546, opened 2022-07-01
LCAM-88

Mhowell494:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-88)

- Fix issue preventing old passport assessments being marked as replaced when one is created
- Enabled ignored integration test that required the above fix to pass.

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEntity.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEntity.java
@@ -109,4 +109,6 @@ public class PassportAssessmentEntity {
     private String whoDWPChecked;
     @Column(name = "RT_CODE")
     private String rtCode;
+    @Transient
+    private Integer financialAssessmentId;
 }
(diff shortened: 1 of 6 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEntity.java
```diff
@@ -109,4 +109,6 @@ public class PassportAssessmentEntity {
     private String whoDWPChecked;
     @Column(name = "RT_CODE")
     private String rtCode;
+    @Transient
+    private Integer financialAssessmentId;
```
mtac50, line 113: Is this being used anywhere?

mtac50: Which mapping is it where its lost?

mtac50: ah I see it, its on the `PassportAssessmentDTO`!

mtac50 (approved):

PR #612, opened 2022-10-31
LCAM-286|Add Entity and Repository classes for Contributions

venkat980:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-286)

Add Entity and Repository classes for Contributions.

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/ContributionsEntity.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/ContributionsEntity.java
@@ -6,6 +6,7 @@ import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.Table;
+import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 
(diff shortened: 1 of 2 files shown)


muthusubra (commented):
What about Impl and Service classes?

venkat980:
> What about Impl and Service classes?

it was decided that we will only create the Entity and Repository classes for now, same as LCAM-287.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/ContributionsEntity.java
```diff
@@ -0,0 +1,97 @@
...
+    @Column(name = "CONT_FILE_ID")
+    private Integer contributionFileId;
+
+    @Column(name = "EFFECTIVE_DATE", nullable = false)
+    private LocalDate effectiveDate;
+
+    @Column(name = "CALC_DATE", nullable = false)
+    private LocalDate calcDate;
+
+    @Column(name = "CONTRIBUTION_CAP", nullable = false)
```
Mhowell494, line 40: This looks like a monetary value, so I think BigDecimal is more appropriate.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/ContributionsEntity.java
```diff
@@ -0,0 +1,97 @@
...
+    private Integer contributionCap;
+
+    @Column(name = "MONTHLY_CONTRIBS", nullable = false)
+    private Integer monthlyContributions;
```
Mhowell494, line 44: This looks like a monetary value, so I think BigDecimal is more appropriate.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/ContributionsEntity.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Column(name = "UPFRONT_CONTRIBS")
+    private Integer upfrontContributions;
```
Mhowell494, line 47: This looks like a monetary value, so I think BigDecimal is more appropriate.

Mhowell494 (changes requested):

Mhowell494 (approved):

PR #693, opened 2023-03-15
clarified few points in cloud platform setup of readme

nigelpain:
## What

Updated cloud platform setup section of readme as I was working through it to clarify a few points and correct a typo or two.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,12 @@ In order to see their actual values you'll need to decode each secret using the 
 echo "<encoded-secret-value>" | base64 --decode
 ```
 
+If you have [jq](https://stedolan.github.io/jq/) installed then you can retrieve the secrets and decode them with just the following command:
+
+```sh
+kubectl -n laa-court-data-adaptor-dev get secret cda-messaging-queues-output -o json | jq '.data|map_values(@base64d)'
+```
+
 Configure AWS details using aws cli command:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,26 +89,34 @@ docker-compose up
...
-kubectl -n laa-court-data-adaptor-dev get secrets ca-messaging-queues-output -o yaml
+kubectl -n laa-court-data-adaptor-dev get secrets cda-messaging-queues-output -o yaml
 ```
 
-You'll need to decode the secrets using the following command. 
+In order to see their actual values you'll need to decode each secret using the following command:
 
 ```sh
-echo 'Your Secret' | base64 --decode
+echo "<encoded-secret-value>" | base64 --decode
```
mtac50, line 108: Another one to add here for anyone who has `jq` installed in their terminal is:
`kubectl -n laa-court-data-adaptor-dev get secret cda-messaging-queues-output -o json | jq '.data|map_values(@base64d)'`
This will retrieve and decode at the same time. :)

nigelpain: Looks like a handy utility! Ive added the above section to the readme for those that have it or want to make use of jq.

nittagr (approved):

mtac50 (approved):

PR #727, opened 2023-05-02
Updated SQS listener from JMS to Spring cloud SQS for Link and unlink

atharmajeed:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/lasb-1926)

Updated the code for linking to manage the duplicate and added a try-catch in the case of a validation error. For any other failure the message will be posted back to the SQS but for validation, it will finish the call. This is the same for both linking and unblinking. 

Upgraded the Unlinking SQS listener from JMS to Spring cloud SQS. 

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/service/UnlinkListener.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/service/UnlinkListener.java
@@ -4,13 +4,17 @@ import com.amazonaws.xray.spring.aop.XRayEnabled;
 import com.google.gson.Gson;
 import gov.uk.courtdata.enums.LoggingData;
 import gov.uk.courtdata.enums.MessageType;
+import gov.uk.courtdata.exception.ValidationException;
 import gov.uk.courtdata.model.Unlink;
 import gov.uk.courtdata.service.QueueMessageLogService;
 import gov.uk.courtdata.unlink.processor.UnLinkProcessor;
+import io.awspring.cloud.messaging.listener.SqsMessageDeletionPolicy;
+import io.awspring.cloud.messaging.listener.annotation.SqsListener;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.slf4j.MDC;
-import org.springframework.jms.annotation.JmsListener;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.handler.annotation.Headers;
 import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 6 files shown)


maat-court-data-api/src/test/java/gov/uk/courtdata/integration/link/service/CreateLinkListenerIntegrationTest.java
```diff
@@ -85,7 +87,11 @@ public void givenSaveAndLinkModel_whenSaveAndImplIsInvoked_thenLinkEstablished()
...
         //when
-        createLinkListener.receive(saveAndLinkMessage, "2");
+        Map<String, Object> header = new HashMap<>();
+        header.put("MessageId", "AIDAIU3GACVJITZULQ2RQ");
+        MessageHeaders headers = new MessageHeaders(header);
+        //when
+        createLinkListener.receive(saveAndLinkMessage, headers);
 
         //then
         CourtDataDTO courtDataDTO = testModelDataBuilder.getSaveAndLinkModelRaw();
```
mtac50, line 97: I think it is worth expanding the link integration tests to explicitly cover a scenario with a validation error. At the moment we only cover the happy path.

maat-court-data-api/src/test/java/gov/uk/courtdata/link/service/CreateLinkListenerTest.java
```diff
@@ -32,7 +35,7 @@ public void givenJSONMessageIsReceived_whenCreateLinkListenerIsInvoked_thenCreat
         String message = "Test JSON";
         //when
         when(gson.fromJson(message, CaseDetails.class)).thenReturn(caseDetails);
-        createLinkListener.receive(message, "1");
+        createLinkListener.receive(message, new MessageHeaders(new HashMap<>()));
```
mtac50, line 38: Same here we only test the happy path so lets add a test case that fails the validation checks.

mtac50 (changes requested):

maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/service/UnlinkListener.java
```diff
@@ -27,13 +31,19 @@ public class UnlinkListener {
...
+                        final @Headers MessageHeaders headers) {
+
+        try {
+            log.debug("message-id {}", headers.get("MessageId"));
+            MDC.put(LoggingData.REQUEST_TYPE.getValue(), MessageType.UNLINK.name());
+            queueMessageLogService.createLog(MessageType.UNLINK, message);
+            Unlink unlink = gson.fromJson(message, Unlink.class);
+            unLinkProcessor.process(unlink);
+        } catch (ValidationException exception) {
+            log.warn(exception.getMessage());
```
lkm0287, line 46: What level are the production logs? For example, if the production logs are at ERROR level we would not log the failing message?

I think there is also a potential security consideration for logging messages in the logs FYI. For example, if a hacker gains entry to the logs could they reverse engineer the payload and obtain some data.

maat-court-data-api/src/test/java/gov/uk/courtdata/integration/link/service/CreateLinkListenerIntegrationTest.java
```diff
@@ -85,7 +87,11 @@ public void givenSaveAndLinkModel_whenSaveAndImplIsInvoked_thenLinkEstablished()
         String saveAndLinkMessage = testModelDataBuilder.getSaveAndLinkString();
 
         //when
-        createLinkListener.receive(saveAndLinkMessage, "2");
+        Map<String, Object> header = new HashMap<>();
+        header.put("MessageId", "AIDAIU3GACVJITZULQ2RQ");
+        MessageHeaders headers = new MessageHeaders(header);
+        //when
+        createLinkListener.receive(saveAndLinkMessage, headers);
```
lkm0287, line 94: I would argue this is not really an integration test as it's not testing SQSListener reading from the message queue.

A local test message queue can be set up and populated with the exact message data to test that this SQSListener is doing as we expect.

We can investigate doing that at another time though - not in the scope of this ticket.

maat-court-data-api/src/main/java/gov/uk/courtdata/unlink/service/UnlinkListener.java
mtac50: This is an interesting point - the log level is only set to INFO at the moment.

mtac50: Ah wait we are receiving WARN and ERROR messages, ignore me!

maat-court-data-api/src/test/java/gov/uk/courtdata/integration/link/service/CreateLinkListenerIntegrationTest.java
mtac50: Yes it would be nice to have a full integration test for this one - @lkm0287 can you add in a spike to the backlog to explore options please?

mtac50 (approved):

PR #763, opened 2023-06-06
LASB-1960 - integration test for AWS SQS

lkm0287:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-1960)

Integration test for SQS.

Basically works with a few caveats:

- Requires a localstack docker image in the local registry.

- A couple of errors in the logs when running the test around SQS credentials (not really sure why) and shutting down the queue that could be looked in at a future date (the queues are actually shut down but looks like it could be done more gracefully).

## Checklist

### DIFF
--- a/maat-court-data-api/src/test/java/gov/uk/courtdata/link/service/CreateLinkListenerEndToEndTest.java
+++ b/maat-court-data-api/src/test/java/gov/uk/courtdata/link/service/CreateLinkListenerEndToEndTest.java
@@ -20,7 +20,7 @@ import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
-public class CreateLinkListenerEndToEndTest extends SqsIntegrationTest {
+class CreateLinkListenerEndToEndTest extends SqsIntegrationTest {
 
     private static final int ONCE = 1;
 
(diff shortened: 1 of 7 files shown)


mtac50:
Nice work @lkm0287! It looks like local stack integrates nicely with CircleCI if we want to use it with any of the other services so @alexdesi could use it without any issues in CDA (assuming there is a Ruby Gem for it...). As for running it in AWS, I think this should also be possible as the base image we use for the pipeline has docker installed. We could put a copy of the local stack image in our AWS container registry and then pull it down before we run the gradle build command in the build.sh script.  We'd have to test it but I think it should work!

maat-court-data-api/src/test/java/gov/uk/courtdata/link/service/CreateLinkListenerEndToEndTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        sqsTemplate.convertAndSend(queueName, messageJson);
+
+        await().atMost(3, SECONDS).untilAsserted(() -> {
+            // assert the SQS is in expected state
+            assertThat(numberOfMessagesInLinkQueue()).isEqualTo(0);
+            assertThat(numberOfMessagesNotVisibleInLinkQueue()).isEqualTo(0);
+
+            // verify behaviour as expected
+            verify(createLinkService, times(ONCE)).saveAndLink(any());
```
skasthuri36326, line 68: Nice work! Is it possible to configure DLQ and test the exception scenarios?

lkm0287: Thanks, I believe you can set a DWQ: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/examples-sqs-dead-letter-queues.html and test it. You also can test the exception scenarios.

mtac50 (approved):

mikereiddigital (approved):
lg2m

PR #816, opened 2023-08-31
LASB-2396: Added additional sqs get queue url permission to the EC2 Role

mtac50:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-2396)

- [x] Added additional sqs permission to the EC2 IAM role.

## Checklist

### DIFF
--- a/aws/application/application.template
+++ b/aws/application/application.template
@@ -440,6 +440,7 @@ Resources:
         Statement:
         - Effect: Allow
           Principal:
+            AWS: [arn:aws:iam::902837325998:role/LAA2ndLineRole]
             Service: [ecs-tasks.amazonaws.com]
           Action: ['sts:AssumeRole']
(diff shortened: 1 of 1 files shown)


aws/application/application.template
```diff
@@ -440,6 +440,7 @@ Resources:
         Statement:
         - Effect: Allow
           Principal:
+            AWS: [arn:aws:iam::902837325998:role/LAA2ndLineRole]
```
mtac50, line 443: Is this correct if I want to allow people with the `LAA2ndLineRole` permission to assume the IAM Role attached to the EC2 Task? I'd like to do this so we can test locally.

mikereiddigital: Worth a try.

mikereiddigital (commented):
lg2m.

PR #829, opened 2023-09-28
LASB-1982 Create RepositoryUtil test utility to standardise clearing down repos correctly after each test

stone-dj:
## What

[[Link to story](https://dsdmoj.atlassian.net/browse/AP-XXX)](https://dsdmoj.atlassian.net/browse/LASB-1982)

As illustrated by https://github.com/ministryofjustice/laa-maat-court-data-api/pull/739 there is a need to provide a standard and reusable way to correctly tear down/delete contents of all repositories in integration tests. This will reduce duplicate and boiler plate code and help ensure that all repositories are in a clean state following any integration test run.

Changes include:

- Add new Utility class and method `gov.uk.courtdata.util.RepositoryUtil.clearUp`
- Replace duplicated `deleteAll()` and `flush()` calls in tests with a single call to `RepositoryUtil.clearUp()`

## Checklist

### DIFF
--- a/maat-court-data-api/src/test/java/gov/uk/courtdata/util/RepositoryUtil.java
+++ b/maat-court-data-api/src/test/java/gov/uk/courtdata/util/RepositoryUtil.java
@@ -1,0 +1,25 @@
+package gov.uk.courtdata.util;
+
+import lombok.experimental.UtilityClass;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.Arrays;
+import java.util.function.Consumer;
+
+@UtilityClass
+public class RepositoryUtil {
+
+    private static final Consumer<JpaRepository<?, ?>> DELETE_ALL_AND_FLUSH = repository -> {
+        repository.deleteAll();
+        repository.flush();
+    };
+
+    /**
+     * The responsibility of this method is to delete all and flush all entities within all the supplied repositories.
+     *
+     * @param repositories The repositories to be cleared
+     */
+    public void clearUp(JpaRepository<?, ?>... repositories) {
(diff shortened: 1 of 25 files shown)


maat-court-data-api/src/test/java/gov/uk/courtdata/util/RepositoryUtil.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.experimental.UtilityClass;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.Arrays;
+import java.util.function.Consumer;
+
+@UtilityClass
+public class RepositoryUtil {
+
+    private static final Consumer<JpaRepository<?, ?>> DELETE_ALL_AND_FLUSH = repository -> {
```
Ross-Nation, line 12: Is the static keyword necessary here? From what I read the UtilityClass annotation makes all members static. :)

lkm0287 (approved):

maat-court-data-api/src/test/java/gov/uk/courtdata/util/RepositoryUtil.java
stone-dj: Yes, agreed, good spot! I have removed the superfluous `static` keyword as it is already implied by `@UtilityClass` annotation. So the `DELETE_ALL_AND_FLUSH` variable now looks like the below.

```
    private final Consumer<JpaRepository<?, ?>> DELETE_ALL_AND_FLUSH = repository -> {
        repository.deleteAll();
        repository.flush();
    };
```

PR #880, opened 2023-11-27
changed the request type from string to Integer.

atharmajeed:
Changed the return type from String to returning a object.

## What

## Checklist

### DIFF
--- a/maat-court-data-api/src/test/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestControllerTest.java
+++ b/maat-court-data-api/src/test/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestControllerTest.java
@@ -51,11 +51,11 @@ class ConcorContributionsRestControllerTest {
                         .contentType(MediaType.APPLICATION_JSON_VALUE))
                 .andExpect(status().isOk())
                 .andExpect(jsonPath("$", hasSize(3)))
-                .andExpect(jsonPath("$.[?(@.concorContributionId==1)].concorContributionId").value(1))
+                .andExpect(jsonPath("$.[?(@.concorContributionId==1)].concorContributionId").exists())
                 .andExpect(jsonPath("$.[?(@.concorContributionId==1)].xmlContent").value("FirstXMLFile"))
-                .andExpect(jsonPath("$.[?(@.concorContributionId==2)].concorContributionId").value(2))
+                .andExpect(jsonPath("$.[?(@.concorContributionId==2)].concorContributionId").exists())
                 .andExpect(jsonPath("$.[?(@.concorContributionId==2)].xmlContent").value("SecondXMLFile"))
-                .andExpect(jsonPath("$.[?(@.concorContributionId==3)].concorContributionId").value(3))
+                .andExpect(jsonPath("$.[?(@.concorContributionId==3)].concorContributionId").exists())
                 .andExpect(jsonPath("$.[?(@.concorContributionId==3)].xmlContent").value("ThirdXMLFile"));
     }
 


maat-court-data-api/src/main/java/gov/uk/courtdata/dces/response/ConcorContributionResponse.java
```diff
@@ -0,0 +1,13 @@
+package gov.uk.courtdata.dces.response;
+
+import lombok.*;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
```
stone-dj, line 8: This class can made immutable as no `set` methods are used and all the fields are immutable (`Integer` and `String`)
You can use the Lombok annotation `@Value` to do this. e.g.
```
@Builder
@Value
public class ConcorContributionResponse {
    Integer concorContributionId;

    @ToString.Exclude
    String xmlContent;
}
```

maat-court-data-api/src/main/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestController.java
```diff
@@ -34,11 +35,11 @@ public class ConcorContributionsRestController {
     @StandardApiResponse
     @GetMapping(value = "/concor-contribution-files", produces = MediaType.APPLICATION_JSON_VALUE)
     @Operation(description = "Get a list of Contributions files")
-    public ResponseEntity<List<String>> getConcorContributionFiles(@RequestParam(name = "status") final ConcorContributionStatus status) {
+    public ResponseEntity<List<ConcorContributionResponse>> getConcorContributionFiles(@RequestParam(name = "status") final ConcorContributionStatus status) {
         log.info("Get Concor contribution files with status {}" ,status);
-        final List<String> concorFiles = concorContributionsService.getConcorFiles(status);
-        log.info("findContributionFiles count {}", concorFiles.size());
-        return ResponseEntity.ok(concorFiles);
+        final List<ConcorContributionResponse> contributionResponses = concorContributionsService.getConcorContributionFiles(status);
```
stone-dj, line 40: I don't see what benefit the `final` modifier provides here? The method is very short (4 lines!) and readable :)
Marking `contributionResponses` as final seems a bit superfluous.

Also the Developer Guidelines as they currently stand do not recommend using the final modifier for local variables.

> Use the final keyword judiciously. Only use final:
> On constants, fields and parameters to indicate immutability
> On methods and classes which absolutely must not be overloaded

https://dsdmoj.atlassian.net/wiki/spaces/ASLST/pages/3896049821/Developer+Guidelines

maat-court-data-api/src/test/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestControllerTest.java
```diff
@@ -38,23 +39,29 @@ class ConcorContributionsRestControllerTest {
...
 
         mvc.perform(MockMvcRequestBuilders.get(String.format(ENDPOINT_URL  +"/concor-contribution-files"))
                         .queryParam("status", ConcorContributionStatus.ACTIVE.name())
                         .contentType(MediaType.APPLICATION_JSON_VALUE))
                 .andExpect(status().isOk())
                 .andExpect(jsonPath("$", hasSize(3)))
-                .andExpect(jsonPath("$[0]").value("FirstXMLFile"))
-                .andExpect(jsonPath("$[1]").value("SecondXMLFile"))
-                .andExpect(jsonPath("$[2]").value("ThirdXMLFile"));
+                .andExpect(jsonPath("$[0].concorContributionId").value("1"))
```
stone-dj, line 53: These assertions are very thorough (which is good!) although making an assumption on the order of the elements in the JSON response make this test a bit brittle and likely to break even if there is no functional change to the endpoint.

You could instead use `JSONAssert`
i.e.
`JSONAssert.assertEquals(expected, actual, JSONCompareMode.NON_EXTENSIBLE);
`

Or change the criteria of the JSON Paths to select elements based on concorContributionId
i.e.
`$[?(@.concorContributionId == 1)]
`

(My preference would be for JSONAssert as it validates the whole of the JSON response and is IMHO easier to read, but the choice is yours)

e.g. This test will fail for the below JSON even though it is functionally equivalent.
```
[
  {
    "concorContributionId": 3,
    "xmlContent": "ThirdXMLFile"
  },
  {
    "concorContributionId": 1,
    "xmlContent": "FirstXMLFile"
  },
  {
    "concorContributionId": 2,
    "xmlContent": "SecondXMLFile"
  }
]
```

stone-dj (changes requested):
I have left some feedback, please action it and let me know when done

maat-court-data-api/src/test/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestControllerTest.java
```diff
@@ -38,23 +40,29 @@ class ConcorContributionsRestControllerTest {
...
 
         mvc.perform(MockMvcRequestBuilders.get(String.format(ENDPOINT_URL  +"/concor-contribution-files"))
                         .queryParam("status", ConcorContributionStatus.ACTIVE.name())
                         .contentType(MediaType.APPLICATION_JSON_VALUE))
                 .andExpect(status().isOk())
                 .andExpect(jsonPath("$", hasSize(3)))
-                .andExpect(jsonPath("$[0]").value("FirstXMLFile"))
-                .andExpect(jsonPath("$[1]").value("SecondXMLFile"))
-                .andExpect(jsonPath("$[2]").value("ThirdXMLFile"));
+                .andExpect(jsonPath("$.[?(@.concorContributionId==1)].concorContributionId").value(1))
```
stone-dj, line 54: For the x3 `andExpect` like this:
`.andExpect(jsonPath("$.[?(@.concorContributionId==1)].concorContributionId").value(1))`

It is redundant to check that the `concorContributionId` value is 1. It is sufficient to verify only that it exists
i.e.
`.andExpect(jsonPath("$.[?(@.concorContributionId==1)]").exists())`

This is because the JSON path `"$.[?(@.concorContributionId==1)].concorContributionId"` basically means "Select all elements which are children of the root element ($.) which have a field named concorContributionId which has a value of 1"
So to then assert that the value of `concorContributionId` is 1 is a bit redundant as we already know that the value must be 1 based on the JSON path above.

stone-dj (changes requested):
Last bit of feedback, changes look good :)

stone-dj (approved):
Changes look good, nice work!

PR #891, opened 2023-12-15
LASB-2804

JHunt16:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-2804)

Added endpoints for POST and DELETE for rep_orders. 

## Checklist

- [x] Github should not be reporting conflicts; you should have recently run `git rebase main`. (I did a merge instead)


maat-court-data-api/src/main/java/gov/uk/courtdata/entity/RepOrderEntity.java
```diff
@@ -51,6 +53,11 @@ public class RepOrderEntity {
     @OneToMany(mappedBy = "repOrder", fetch = FetchType.LAZY, cascade = CascadeType.ALL, orphanRemoval = true)
     private final List<IOJAppealEntity> iojAppeal = new ArrayList<>();
 
+    @ToString.Exclude
+    @JsonIgnore
+    @OneToMany(mappedBy = "repOrder", fetch = FetchType.LAZY, cascade = CascadeType.REMOVE, orphanRemoval = true)
```
lkm0287, line 58: What does this cascade type do?
Also, why have we got this JsonIgnore and the other don't?

maat-court-data-api/src/main/java/gov/uk/courtdata/model/CreateRepOrder.java
```diff
@@ -0,0 +1,86 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+@Data
+@SuperBuilder
```
lkm0287, line 12: As this class doesn't extend anything I don't think the SuperBuilder does anything?

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
```diff
@@ -51,6 +54,17 @@ public RepOrderDTO updateDateCompleted(final UpdateAppDateCompleted updateAppDat
...
 
+    @Transactional
+    public RepOrderDTO create(final CreateRepOrder createRepOrder) {
+        log.info("create rep order - Transaction Processing - Start");
+        LocalDate dateNow = LocalDate.now();
+        RepOrderEntity repOrderEntity = new RepOrderEntity();
+        repOrderEntity.setId(createRepOrder.getRepId());
+        repOrderEntity.setDateCreated(dateNow);
+        repOrderMapper.createRepOrderToRepOrderEntity(createRepOrder, repOrderEntity);
+        return repOrderMapper.repOrderEntityToRepOrderDTO(repOrderImpl.createRepOrder(repOrderEntity));
```
lkm0287, line 65: The repOrderImpl.createRepOrder(repOrderEntity) is actually doing the saving of the entity so my recommendation is to put it on its own line rather than passed as a param here.

The reason for this is if there are any errors thrown during saving the entity it'll show the correct line in the stacktrace (it might confusingly show this repOrderMapper line).

maat-court-data-api/src/main/java/gov/uk/courtdata/reporderhistory/entity/RepOrderHistoryEntity.java
```diff
@@ -0,0 +1,240 @@
...
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "REP_ID", nullable = false, updatable = false)
+    private RepOrderEntity repOrder;
+
+    @Id
+    @Column(name = "ID")
+    private Integer id;
+
+    @Column(name = "CASE_ID")
+    private String caseId;
```
lkm0287, line 31: I had a quick look in the dev DB on this table and there are constraints such as this field being not null on the table. It is usually helpful to catch these type of errors by having the constraints here in the entity before the DB throwing them.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/RepOrderEntity.java
JHunt16: The cascade ensures the RepOrderHistory is deleted for this RepOrder when the RepOrder is deleted.

I've removed the JsonIgnore - I forgot to remove it after earlier issues and experimentation :)

maat-court-data-api/src/main/java/gov/uk/courtdata/model/CreateRepOrder.java
JHunt16: I've replaced this with a normal @Builder now.

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
JHunt16: Makes sense, done.

maat-court-data-api/src/main/java/gov/uk/courtdata/validator/MaatIdValidator.java
```diff
@@ -38,4 +41,21 @@ public Optional<Void> validate(final Integer maatId) {
...
+    /**
+     * @param maatId
+     * @return
+     * @throws ValidationException
+     */
+    public void validateNotExists(Integer maatId) {
+
+        if (maatId != null && maatId > 0) {
+            if (repOrderService.exists(maatId)) {
+                throw new ValidationException(String.format("There is already a record with this MAAT ID."));
```
stone-dj, line 54: Should include the value of the `maatId` in the exception message

e.g.
`"There is already a record with MAAT ID [%d]".formatted(maatId)`

maat-court-data-api/src/main/java/gov/uk/courtdata/reporderhistory/entity/RepOrderHistoryEntity.java
JHunt16: I've not added them to this particular entity because I only created this entity for the purposes of cascade deleting when deleting a rep order (I need this new RepOrderHistoryEntity to do the delete). We don't do any creation or updating of the rep_orders_history table within this application - this all happens in DB triggers within stored procedures in MAAT, so don't need the validation on the entity at the moment.

However, I have now added @NotNull annotations to the required fields on my CreateRepOrder class, so they are now caught as a BadRequest if not submitted with the POST request.

maat-court-data-api/src/main/java/gov/uk/courtdata/validator/MaatIdValidator.java
```diff
@@ -38,4 +41,21 @@ public Optional<Void> validate(final Integer maatId) {
         }
 
     }
```
stone-dj, line 43: **A nice to have** Could you rename the existing method to something more descriptive?

i.e. 
`    public Optional<Void> validate(final Integer maatId) {`
to
`    public Optional<Void> validateExists(final Integer maatId) {
`

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/mapper/RepOrderMapper.java
```diff
@@ -10,5 +11,7 @@ public interface RepOrderMapper {
 
     RepOrderDTO repOrderEntityToRepOrderDTO(final RepOrderEntity repOrder);
 
+    void createRepOrderToRepOrderEntity(CreateRepOrder createRepOrder, @MappingTarget RepOrderEntity repOrderEntity);
```
stone-dj, line 14: Good use of `mapstruct` :)

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
```diff
@@ -145,6 +147,30 @@ public ResponseEntity<Object> findByRepIdAndVehicleOwner(@PathVariable int repId
         ));
     }
 
+    @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Create a rep order record")
+    @ApiResponse(responseCode = "200",
```
stone-dj, line 153: Many of these duplicated boiler plate annotations can be replaced with 
[StandardApiResponse.java](https://github.com/ministryofjustice/laa-maat-court-data-api/blob/9bade7c6c2b82ec82067c67074c7c890272690e8/maat-court-data-api/src/main/java/gov/uk/courtdata/annotation/StandardApiResponse.java)

See [ConcorContributionsRestController.java](https://github.com/ministryofjustice/laa-maat-court-data-api/blob/0e4ca828653a95453747bc1e6fedb2b2988a7a1b/maat-court-data-api/src/main/java/gov/uk/courtdata/dces/controller/ConcorContributionsRestController.java) for an example

stone-dj (changes requested):
Have left some review feedback, changes look good

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/RepOrderEntity.java
lkm0287: Thanks!

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
JHunt16: I've kept in the 200 response annotations here, as they appear to be unique to each of the methods (with their own operation description). However I've replaced the 400 and 500 ApiResponses in this controller with the StandardApiResponse annotation to reduce some of that boilerplate.

maat-court-data-api/src/main/java/gov/uk/courtdata/validator/MaatIdValidator.java
JHunt16: I've decided to leave it as-is because there are several other validator classes within the same folder that are structured in the same way (with a single validate method), so have tried to keep as consistent as I can. I may have validated single responsibility by adding the 'validateNotExists' method to this class, but I feel it makes sense to allow these classes to do a bit more than just have a single validate method. There may be opportunities for some refactoring here on some of these other classes.

stone-dj: yeah no worries, makes sense :)

PR #905, opened 2024-01-16
LASB-2805 Add endpoints to allow CRUD functionality on the rep_order_equity table

JHunt16:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-2805)

Added endpoints to allow CRUD functionality on the rep_order_equity table.

## Checklist

- [x] Github should not be reporting conflicts; you should have recently run `git rebase main`. **(Merged main into this feature branch instead)**
- [] You should have checked that the commit messages say why the change was made.


maat-court-data-api/src/main/java/gov/uk/courtdata/entity/RepOrderEquityEntity.java
```diff
@@ -0,0 +1,88 @@
...
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "rep_order_equity_seq")
+    private Integer id;
+
+    @NotNull
+    @Column(name = "REP_ID")
+    private Integer repId;
+
+    @Nullable
+    @Column(name = "ROPD_ID")
+    private Integer ropdId;
```
Ross-Nation, line 36: Do we know what ropd stands for? Wonder if we can make our variable name more descriptive even if the column name is not. :)

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderEquityController.java
```diff
@@ -0,0 +1,46 @@
...
+
+@RestController
+@RequiredArgsConstructor
+@RequestMapping("/api/internal/v1/assessment/rep-order-equity")
+public class RepOrderEquityController {
+
+    private final RepOrderEquityService repOrderEquityService;
+
+    @GetMapping(value = "/{id}")
+    @StandardApiResponseCodes
```
Ross-Nation, line 18: So, this was an interesting learn for me around the custom annotations! 

Firstly, you're pulling this annotation from within the 'eform.controller' package, so I think in the interested of keeping things in the correct areas, if you want to use it then it's probably worth either creating a new 'StandardApiResponseCodes' within the 'reporder.controller' package or making the package more 'global' by moving it out of eform.controller and into the higher level courtdata.controller area... Alex might have some thoughts on this second way of doing things though. 

It's also worth adding, that this doesn't look to cover the '404' not found response, so you might want to add that into the get and update calls. As far as I understand this is just 'documentation' and doesn't impact functionality, but it would be good to ensure it's accurate. :)

Ross-Nation: The ResultController class is a good place to look at this. Also, it makes use of @operation too, which is basically just a description of what you expect the endpoint to do.

JHunt16: It looks like there is a StandardApiResponse annotation instead, which appears to be global and not eform-specific, so have switched to that. It doesn't include 200 responses though for some reason, so have added them separately, along with 404 annotations where appropriate.

Also added the operation and tag annotations to make the swagger docs more descriptive here.

PR #916, opened 2024-01-29
CRUD for Applicants table

muthusubra:
## What

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/applicant/controller/ApplicantController.java
```diff
@@ -61,4 +65,42 @@ public ResponseEntity<ApplicantHistoryDTO> updateApplicantHistory(@RequestBody @
...
+        log.info("Get Applicant Request Received");
+        return ResponseEntity.ok(applicantService.find(id));
+    }
+
+    @PatchMapping(value = "/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Update a Applicant record")
+    @StandardApiResponseCodes
+    @NotFoundApiResponse
+    public ResponseEntity<Void> updateApplicant(@PathVariable int id, @RequestBody @Valid ApplicantDTO applicantDTO) {
+        log.info("Update Applicant Request Received");
```
skasthuri36326, line 84: We need to discuss on patch operation. We need to use the @DynamicUpdate at entity level.

muthusubra: As discussed, will not be using DynamicUpdate for this use case. Resolving the comment

venkat980 (approved):

PR #924, opened 2024-02-05
Lcam 1215

muthusubra:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1215)

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/applicant/controller/ApplicantController.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/applicant/controller/ApplicantController.java
@@ -80,7 +80,7 @@ public class ApplicantController {
     @PostMapping(value = "/applicant-history", produces = MediaType.APPLICATION_JSON_VALUE)
     @Operation(description = "Create a Applicant History record")
     @StandardApiResponseCodes
-    public ResponseEntity<Applicant> createApplicantHistory(@RequestBody @Valid ApplicantHistoryEntity applicantHistoryEntity) {
+    public ResponseEntity<ApplicantHistoryEntity> createApplicantHistory(@RequestBody @Valid ApplicantHistoryEntity applicantHistoryEntity) {
         log.info("Create Applicant Request Received");
         applicantHistoryService.create(applicantHistoryEntity);
         return ResponseEntity.ok().build();


maat-court-data-api/src/main/java/gov/uk/courtdata/applicant/controller/ApplicantController.java
```diff
@@ -67,6 +68,24 @@ public ResponseEntity<ApplicantHistoryDTO> updateApplicantHistory(@RequestBody @
...
+    public ResponseEntity<Void> deleteApplicantHistory(@PathVariable int id) {
+        log.info("Delete Applicant Request Received");
+        applicantHistoryService.delete(id);
+        return ResponseEntity.ok().build();
+    }
+
+    @PostMapping(value = "/applicant-history", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "Create a Applicant History record")
+    @StandardApiResponseCodes
+    public ResponseEntity<Applicant> createApplicantHistory(@RequestBody @Valid ApplicantHistoryEntity applicantHistoryEntity) {
```
skasthuri36326, line 83: We usually pass DTO to the controller as the Entity exposes some of the attributes that the user should not set like dateCreated, dateModified. Any reason ?

muthusubra: In this case it is ‘create’ and the data is passed in. 

Even in other cases, between micro services call, the consumers are internal consumers, and there is no concern of potentially exposing sensitive information. 

Also, in one of earlier conversation it was agreed that we will not use DTO for the sake of using it and we will use it only if the client interface data object is different from the underlying entity and if the DTO makes communication between layers more efficient. We should not be using DTO just as a pass through and it will add to the template code and will make the call more inefficient

skasthuri36326: Ok, the existing get and update endpoints for applicant history will be used in the pre update checks and it uses the DTO. Could you please change those endpoints to use entity to keep it consistent or else it breaks the data contract?

muthusubra: Given that bot the DTO and the Entity are exactly the same, as it is already implemented, I think we can leave it as it is

PR #929, opened 2024-02-13
DCES-Adding-Maat-Id-To-Fdc-Get

LAA-Dave-Edwards:
## What

Adding maat id to the fdc get call
Adding rep order population and testing of the id.

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
@@ -11,6 +11,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Function;
 
 @Slf4j
 @Service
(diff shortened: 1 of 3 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
```diff
@@ -24,6 +24,7 @@ public FdcContributionsResponse getFdcContributionFiles(FdcContributionsStatus s
 
         List<FdcContributionEntry> fdcContributionEntries = fdcFileList.stream().map(cc -> FdcContributionEntry.builder()
                         .id(cc.getId())
+                        .maatId(cc.getRepOrderEntity().getId())
```
stone-dj, line 27: Just an FYI, sometimes it can be helpful...
either for expressiveness via naming,
or helping to ensure that you are not using any potentially non-thread safe object state,
or being closer to using streams as 'pure' functional constructs...
To define functions as static variables e.g.

```
    private static final Function<FdcContributionsEntity, FdcContributionEntry> TO_FDC_CONTRIBUTION_ENTRY = cc ->
            FdcContributionEntry.builder()
                    .id(cc.getId())
                    .maatId(cc.getRepOrderEntity().getId())
                    .finalCost(cc.getFinalCost())
                    .dateCalculated(cc.getDateCalculated())
                    .lgfsCost(cc.getLgfsCost())
                    .agfsCost(cc.getAgfsCost())
                    .sentenceOrderDate(Objects.nonNull(cc.getRepOrderEntity()) ? cc.getRepOrderEntity().getSentenceOrderDate() : null)
                    .build();
....
        List<FdcContributionEntry> fdcContributionEntries = fdcFileList.stream()
                .map(TO_FDC_CONTRIBUTION_ENTRY)
                .toList();
```

LAA-Dave-Edwards: Quite agree, done.

stone-dj (approved):
Changes look good

PR #951, opened 2024-04-17
LASB-3216 Enable Open API/Swagger documentation endpoint in non-prod environments

stone-dj:
## Enable Open API/Swagger documentation endpoint in non-prod environments

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-3216)

Expose the Open API/Swagger documentation endpoint in all non-prod environments to allow developers and non-Java developers etc. to discover and explore the endpoints available in the MAAT Court Data API.

Changes include:
- Enable unauthenticated access to the endpoint `/open-api/*` in all non-prod environments via a new `AWS::ApiGatewayV2::Route` called `ApiRouteOpenAPI`.

## Checklist

### DIFF
--- a/aws/application/application.template
+++ b/aws/application/application.template
@@ -613,7 +613,7 @@ Resources:
       Description: !Sub '${pAppName} API Gateway - HTTP API'
       ProtocolType: HTTP
   ApiIntegration:
-    Type: 'AWS::ApiGatewayV2::Integration'
+    Type: AWS::ApiGatewayV2::Integration
     Properties:
       ApiId: !Ref ApiGateway
       Description: !Sub '${pAppName} Integration Proxy'
@@ -623,8 +623,8 @@ Resources:
       ConnectionType: VPC_LINK
       ConnectionId: !Ref ApiGatewayVpcLink
       PayloadFormatVersion: '1.0'
-  ApiRoute1:
-    Type: 'AWS::ApiGatewayV2::Route'
+  ApiRouteLinkValidate:
+    Type: AWS::ApiGatewayV2::Route
     DependsOn:
       - ApiIntegration
     Properties:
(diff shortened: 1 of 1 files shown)


nittagr (approved):

lkm0287:
The only thing I'd have to say about this - have you checked we have a firewall with a whitelist for this service? 

We would only want say dev's through global connect to access our Swagger.

I know we've only opened up the Swagger on Cloud Platform after implementing firewall IP whitelists.

stone-dj:
> The only thing I'd have to say about this - have you checked we have a firewall with an allow list (whitelist 😬) for this service?
> 
> We would only want say dev's through global connect to access our Swagger.
> 
> I know we've only opened up the Swagger on Cloud Platform after implementing firewall IP allowlists.

I've not checked any firewall rules etc. I would imagine the group of people who would have access to the newly opened up swagger endpoint would be same group of people who can already access all our API endpoints.
So assuming our current firewall rules etc. are adequate I don't see any issue, I would be very surprised if our current setup has MAAT Court Data API endpoints open to the public internet relying only on authentication via OAuth to secure them...

lkm0287:
> > The only thing I'd have to say about this - have you checked we have a firewall with an allow list (whitelist 😬) for this service?
> > We would only want say dev's through global connect to access our Swagger.
> > I know we've only opened up the Swagger on Cloud Platform after implementing firewall IP allowlists.
> 
> I've not checked any firewall rules etc. I would imagine the group of people who would have access to the newly opened up swagger endpoint would be same group of people who can already access all our API endpoints. So assuming our current firewall rules etc. are adequate I don't see any issue, I would be very surprised if our current setup has MAAT Court Data API endpoints open to the public internet relying only on authentication via OAuth to secure them...

Swagger is a common attach vector in pen tests that we frequently have. At the moment external agents don't know our APIs. If we expose Swagger we give them attack surface to probe us. I'll just check a minute to see if we have IP allow lists.

vc13837 (approved):

PR #962, opened 2024-05-09
LASB-3248 LaatransactionId not matching the MaatId in logs - Fixes for setting and clearing down FDC

stone-dj:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-3248?atlOrigin=eyJpIjoiNDQ1ZWY5NWEwNzQzNGJiYWEzMTg1MWU4YWFhZDUxZjIiLCJwIjoiaiJ9)

Fixed bug of LaatransactionId not matching the MaatId

To ensure that the MDC (Mapped Diagnostic Context) is accurate and contains the correct data relevant to the thread (i.e. Maat Id and LAA Transaction Id) we need to:
- Clear down/wipe the MDC after every request has been processed.
- Set the MAAT Id and LAA Transaction Id once and only once for each session.
- Ensure MDC data is set consistently.
- Update `SqsGlobalLoggingHandler` to ensure the MDC is cleared after processing SQS messages.

## Checklist


maat-court-data-api/src/main/java/gov/uk/courtdata/controller/AfterResponseClearDown.java
```diff
@@ -0,0 +1,30 @@
...
+    
+    @Override
+    public void doFilter(ServletRequest servletRequest,
+        ServletResponse servletResponse,
+                         FilterChain filterChain) throws IOException, ServletException {
+        try {
+            filterChain.doFilter(servletRequest, servletResponse);
+        } finally {
+            MDC.clear();
+            log.debug("Cleared down the MDC");
```
lkm0287, line 27: Filters run in an order (I'm sure you know) - I'm not exact what the default order is however we'd want to make sure that the ordering in the chain doesn't run this MDC.clear() before the filter chain is completed.

mtac50: I think by calling `MDC.clear()` after the `filterChain.doFilter(servletRequest, servletResponse)` you are essentially forcing it to run after the response has been generated. I guess we could also set the filter order to `HIGHEST_PRECEDENCE`  that way it would ensure that the `MDC.clear()` is called after the entire filter chain (request and response) has been completed?

maat-court-data-api/src/main/java/gov/uk/courtdata/exception/GlobalAppLoggingHandler.java
```diff
@@ -68,15 +70,21 @@ public void beforeQueueMessageRec(JoinPoint joinPoint, String message) {
         }
 
         Sentry.configureScope(scope -> {
-            scope.setTag(LoggingData.CASE_URN.getValue(), laaTransactionLogging.getCaseUrn() != null ? laaTransactionLogging.getCaseUrn() : "");
-            scope.setTag(LoggingData.LAA_TRANSACTION_ID.getValue(), laaTransactionIdStr);
-            scope.setTag(LoggingData.MAATID.getValue(), laaTransactionLogging.getMaatId().toString());
+            scope.setTag(LoggingData.CASE_URN.getKey(),
```
mtac50, line 73: Do we also need to be clearing out the MDC in the `afterProcessEnds` method for completeness?

maat-court-data-api/src/main/java/gov/uk/courtdata/controller/AfterResponseClearDown.java
```diff
@@ -0,0 +1,33 @@
...
+        ServletResponse servletResponse,
+                         FilterChain filterChain) throws IOException, ServletException {
+        try {
+            filterChain.doFilter(servletRequest, servletResponse);
+        } finally {
+            log.debug("About to clear down the MDC");
+            MDC.clear();
+        }
+    }
+}
```
lkm0287, line 33: This filter runs last during the request processing and first during the response processing. If any of the other filters uses the MDC in their finally blocks it's not going to log properly _I think_.

lkm0287 (approved):

PR #988, opened 2024-06-18
LCAM-1390|Implement exception handling in StoredProcedure repository

venkat980:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1390)

Implemented the exception handling in the StoredProcedureRepository to be inline with the MAAT application exception handling of the stored procedure calls.

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/repository/StoredProcedureRepository.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/repository/StoredProcedureRepository.java
@@ -18,7 +18,6 @@ import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.HashMap;
 
-
 @Slf4j
 @Repository
 public class StoredProcedureRepository {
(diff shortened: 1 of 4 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/repository/StoredProcedureRepository.java
```diff
@@ -170,15 +170,34 @@ private boolean connectionRequiresManualCommit(Connection connection) throws SQL
...
+                //remove stack trace codes from exception
+                int start = errorMessage.indexOf(':');
+                int end = errorMessage.indexOf('\n');
+                if (start > 0 && end > 0) {
+                    errorMessage = "Application Error " + errorMessage.substring(start, end);
+                }
+                throw new MAATApplicationException(errorMessage, sqlException);
+            }
+        }
+    }
```
nigelpain, line 201: Just a curious question, why the need to only handle the sql errors in the min - max range? If I understand right anything out of this range will still be handled by the default handler and returned as 500 with the exception message, just these have the message tidied up a bit?

nigelpain (approved):

maat-court-data-api/src/main/java/gov/uk/courtdata/repository/StoredProcedureRepository.java
venkat980: Good question, so this specific range of  SQL error codes are being used to raise application errors in the stored procedure, such as validation errors, that need to be displayed on the UI. The application errors are identified with new error code, all other DB errors the generic error message will be displayed.

PR #989, opened 2024-06-19
LASB-3412 - add new endpoint to find MAAT IDs by USN

lkm0287:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-3412)

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
@@ -27,6 +27,7 @@ import jakarta.validation.Valid;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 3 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
```diff
@@ -90,6 +97,17 @@ public ResponseEntity<RepOrderDTO> updateApplicationDateCompleted(@RequestBody U
         return ResponseEntity.ok(repOrderService.updateDateCompleted(updateAppDateCompleted));
     }
 
+    @GetMapping(value = "/rep-order-usn/{usn}", produces = MediaType.APPLICATION_JSON_VALUE)
```
mtac50, line 100: @lkm0287 as the resource being accessed is still a `rep-order` could we not just use a `@RequestParam` so the endpoint would be `/rep-orders?usn=x` instead of `/rep-order/rep-orders-usn/x`?

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
```diff
@@ -90,6 +97,17 @@ public ResponseEntity<RepOrderDTO> updateApplicationDateCompleted(@RequestBody U
...
+    @Operation(description = "Retrieve rep order records by USN")
+    @ApiResponse(responseCode = "200",
+            content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE)
+    )
+    @StandardApiResponse
+    public ResponseEntity<Object> findByUsn(@PathVariable int usn) {
```
stone-dj, line 106: If I have understood it correctly, it will return a list of associated MAAT Ids.
So the response type should probably be `Set<Long>`, or maybe `List<Long>`?

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
```diff
@@ -90,6 +97,17 @@ public ResponseEntity<RepOrderDTO> updateApplicationDateCompleted(@RequestBody U
...
+        log.debug("Get Rep Order By USN Received");
+        List<RepOrderEntity> repOrderEntityList = repOrderRepository.findByUsn(usn);
+        return ResponseEntity.ok(repOrderEntityList.stream().map(RepOrderEntity::getId).collect(Collectors.toSet()));
```
stone-dj, line 109: Good use of `collect(Collectors.toSet())` 👍

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
lkm0287: The response entity still has a useful status field which are used on the functional test end hence wrapped in a ResponseEntity.

stone-dj: discussed, happy for you to make a call on this one 👍

stone-dj (approved):
Changes look good

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/RepOrderController.java
lkm0287: Thanks, agree :) updated.

PR #1011, opened 2024-07-15
LASB-3457: Updated permission for DCES client to allow access to rep order data

mtac50:
# What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-3457)

- [x] Added a new DCES standard scope and added DCES clients to the standard `ApiAuthorizer` audience.

## Checklist

### DIFF
--- a/aws/application/application.template
+++ b/aws/application/application.template
@@ -94,6 +94,10 @@ Parameters:
     Type: String
     Default: read
     Description: Name of read-only Cognito scope
+  pApiScopeDcesStandard:
+    Type: String
+    Default: dces-standard
+    Description: Name of read-only Cognito scope
   pApiStageName:
     Type: String
     Default: "v1"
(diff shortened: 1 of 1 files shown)


aws/application/application.template
```diff
@@ -700,6 +704,8 @@ Resources:
           - !Ref CognitoUserPoolClaimForCrownCourtDefenceClient
           - !Ref CognitoUserPoolClientCrimeValidationService
           - !Ref CognitoUserPoolClientATSService
+          - !Ref CognitoUserPoolClientDcesReportService
+          - !Ref CognitoUserPoolClientDcesDrcReportService
```
nittagr, line 708: I think we have to recreate a new Cognito credential to pick up the new scope for DCES services. Please could you verify?

aws/application/application.template
```diff
@@ -894,7 +900,9 @@ Resources:
       GenerateSecret: true
       SupportedIdentityProviders: [ COGNITO ]
       AllowedOAuthFlows: [ client_credentials ]
-      AllowedOAuthScopes: [ !Sub '${pAppName}/${pApiScope1}' ]
+      AllowedOAuthScopes:
+        - !Sub '${pAppName}/${pApiScopeDcesStandard}'
+        - !If [ cNonProd, !Sub '${pAppName}/${pApiScope1}', !Ref AWS::NoValue ]
```
nittagr, line 905: We need to add a new scope for CognitoUserPoolClientDcesDrcReportService service as well same as CognitoUserPoolClientDcesReportService.

Do we need to add new scope to CognitoUserPoolResourceServer1, please could you verify :-)

mtac50: Ah yes good spot!

aws/application/application.template
mtac50: I remember you saying about this before! As long as the resource server is created first then the client should know the scope exists. I will add a dependency to ensure the creation order.

mtac50: ah we already have the dependency - we should be good!

PR #1021, opened 2024-07-29
Define all dependency and plugin versions in `git.properties`

stone-dj:
## Define all dependency and plugin versions in `git.properties`

This will allow us to quickly and efficiently address **synk** dependency issues.

Changers include:
- Create new `git.properties` file in which we will define all dependency and plugin versions.
- Create new `cleanBuildAndRun_Local.sh` to provide a single command to clean, build and start the application.
- Modify `maat-court-data-api/build.gradle` to reference version numbers in `git.properties`.

Example screenshots:

Screenshot 2024-07-30 at 09 59 39
...
Screenshot 2024-07-30 at 10 04 21

## Checklist


Ross-Nation (approved):
I approve this message!

PR #1027, opened 2024-08-08
LCAM 1505 Illegal Argument Exception Finding Hardship Details for Solicitors Costs

nigelpain:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/projects/LCAM/boards/881?selectedIssue=LCAM-1505)

Added method to retrieve HardshipReviewDetailType enum value based on String provided in request.

## Checklist


karlbaker02:
Looks good! Two questions:

- when we call `HardshipReviewDetailType.getFrom(detailType)`, do we have certainty around the casing of `detailType` or would it be useful to do a case-insensitive match in the `getFrom` method?
- do you think it's worth creating a new test around the existing _HardshipReviewDetailType_ enum (which currently has no tests)?

nigelpain:
> Looks good! Two questions:
> 
> * when we call `HardshipReviewDetailType.getFrom(detailType)`, do we have certainty around the casing of `detailType` or would it be useful to do a case-insensitive match in the `getFrom` method?
> * do you think it's worth creating a new test around the existing _HardshipReviewDetailType_ enum (which currently has no tests)?

- Pretty sure I've only seen the detailType specified all in uppercase, though it's not much to just convert to upper before matching in that method, so I'll stick it in for an extra level of certainty.
- Good point about the unit tests aswell we have them for other enums so I'll add a set for this HardshipReviewDetailType aswell thanks 👍

venkat980 (commented):
We have moved all the Enums along with the tests to common-classes, should we use the enum from the common-classes jar instead of duplicating the tests in MAAT-API?

nigelpain:
> We have moved all the Enums along with the tests to common-classes, should we use the enum from the common-classes jar instead of duplicating the tests in MAAT-API?

Very good point! I have remove the enum from maat-api and updated the imports to point to the crime-commons version of it 👍

venkat980 (approved):

karlbaker02:
One more question: are we using any of the methods of the base `AbstractEnumConverter` for the enum that has now been removed? I notice that the same-named enum in crime-commons doesn't have the majority of the annotations the one previously in this repo does.

nigelpain:
> One more question: are we using any of the methods of the base `AbstractEnumConverter` for the enum that has now been removed? I notice that the same-named enum in crime-commons doesn't have the majority of the annotations the one previously in this repo does.

No we aren't was discussing this with Venkat earlier this morning and we decided to take a different approach for the enum conversion by switch the type in the entity to a String and then implementing a custom conversion method in the HardshipDetailMapper to map the String to the enum in the DTO, rather than implementing the AbstractEnumConverter in the commons library.

PR #1036, opened 2024-08-20
LCAM 1528 Hardship Details Not Being Set to Active on Database

nigelpain:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/projects/LCAM/boards/881?selectedIssue=LCAM-1528)

- Allowed active for HardshipReviewDetail to be deserialised so that we can pass a value of Y through from the hardship service when creating a hardship review.

## Checklist


stone-dj (approved):
Change looks good

PR #1040, opened 2024-08-27
LCAM-1539

Mhowell494:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1539)

- Update default value for `active` field in the `HardshipReviewProgressEntity` to `'Y'`
  - Previously, the active field would default to `'N'`, however, this meant progress items were not being displayed in the UI after creation. New records should always be marked as `active` and when performing an update old progress items are deleted and recreated rather than updating the existing rows.


karlbaker02 (approved):
Looks good 👍

PR #1046, opened 2024-09-05
Add feature flag memberships to user summary information

karlbaker02:
This PR adds feature flag memberships to the user summary information, both for user-specific memberships and for those which apply to all users.

This update is made to allow for changes in services which use feature flags to control logic and specifically, to remove the need to pass feature flag information between services (such as between MAAT and the Orchestration Service). The feature flag information is added specifically to the `UserSummaryDTO` as this is the most appropriate place to hold this information, similar to that held already for role data items.

[Part of LCAM-1342](https://dsdmoj.atlassian.net/browse/LCAM-1342)

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FeatureToggleEntity.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FeatureToggleEntity.java
@@ -1,0 +1,32 @@
+package gov.uk.courtdata.entity;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Id;
+import jakarta.persistence.Table;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity
+@Data
+@Table(name = "FEATURE_TOGGLE", schema = "TOGDATA")
+public class FeatureToggleEntity {
+
+  @Id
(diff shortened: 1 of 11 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FeatureToggleEntity.java
```diff
@@ -0,0 +1,32 @@
...
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+@Entity
+@Data
+@Table(name = "FEATURE_TOGGLE", schema = "TOGDATA")
+public class FeatureToggleEntity {
+
+  @Id
+  @Column(name = "ID", nullable = false)
```
venkat980, line 21: Looking at the FEATURE_TOGGLE table, ID uses a sequence. Should we keep the entity inline with the DB table please?

venkat980 (commented):
Looks good to me, just one comment.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FeatureToggleEntity.java
karlbaker02: Done. I'm not sure if the sequence name here needs to match one that already exists or not - I couldn't see anything in the db. If you could confirm that'd be great, thanks.

venkat980: Looking at the DB, the sequence we are using is ISEQ$$_316195

venkat980 (approved):

PR #1047, opened 2024-09-09
Lasb 3580/get reporder by repid

Ross-Nation:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-3580)

Added an endpoint to the RepOrderController to return rep order state details. Tested locally with postman and all running as expected. 

## Checklist

### DIFF
--- a/maat-court-data-api/src/test/java/gov/uk/courtdata/reporder/controller/RepOrderControllerTest.java
+++ b/maat-court-data-api/src/test/java/gov/uk/courtdata/reporder/controller/RepOrderControllerTest.java
@@ -411,6 +411,66 @@ class RepOrderControllerTest {
     }
 
     @Test
+    void givenValidRepId_whenFindRepOrderStateByRepIdIsInvokedAndFindsARepOrder_thenCorrectResponseIsReturned() throws Exception {
+        RepOrderStateDTO repOrderStateDTO = RepOrderStateDTO.builder()
+            .usn(USN_VALUE)
+            .maatRef(MAAT_REF_VALUE)
+            .caseId(CASE_ID_VALUE)
+            .caseType(CASE_TYPE_VALUE)
+            .iojResult(IOJ_RESULT_VALUE)
+            .iojAssessorName(new AssessorDetails(IOJ_ASSESSOR_FULL_NAME, IOJ_ASSESSOR_USERNAME))
+            .dateAppCreated(LocalDate.parse(DATE_APP_CREATED_VALUE))
+            .iojReason(null)
+            .meansInitResult(MEANS_INIT_RESULT_VALUE)
+            .meansInitStatus(MEANS_INIT_STATUS_VALUE)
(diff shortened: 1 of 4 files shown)


maat-court-data-api/src/test/java/gov/uk/courtdata/reporder/controller/RepOrderControllerTest.java
```diff
@@ -410,6 +390,46 @@ void givenValidUsn_whenFindRepOrderByUsnIsInvokedAndFindsNoRepOrder_thenCorrectR
...
+            .andExpect(jsonPath("$.caseType").value(CASE_TYPE_VALUE))
+            .andExpect(jsonPath("$.iojResult").value(IOJ_RESULT_VALUE))
+            .andExpect(jsonPath("$.iojAssessorName.fullName").value(IOJ_ASSESSOR_FULL_NAME))
+            .andExpect(jsonPath("$.iojAssessorName.userName").value(IOJ_ASSESSOR_USERNAME))
+            .andExpect(jsonPath("$.dateAppCreated").value(DATE_APP_CREATED_VALUE))
+            .andExpect(jsonPath("$.iojReason").doesNotExist())
+            .andExpect(jsonPath("$.meansInitResult").value(MEANS_INIT_RESULT_VALUE))
+            .andExpect(jsonPath("$.meansInitStatus").value(MEANS_INIT_STATUS_VALUE))
+            .andExpect(jsonPath("$.meansFullResult").doesNotExist())
+            .andExpect(jsonPath("$.meansFullStatus").doesNotExist())
```
JHunt16, line 413: Do we need a test for the state of a full means assessment?

Ross-Nation: Good question. I'm ngl, I ripped this off of the USN endpoint tests so maybe didn't think enough about it. Let me have a look. 

I know the mapping of all fields is tested in the mapper tests and I did also test this manually so I know it works... but I will think about it and get back to you!

Ross-Nation: So I had a look at maybe creating a few different mapping scenarios to cover each of the fields but looking at the mapping tests, this is already done there. So I'm leaning on the side of this being unnecessary duplication to do it in the controller tests as well. 

Mapper tests are here: https://github.com/ministryofjustice/laa-maat-court-data-api/blob/main/maat-court-data-api/src/test/java/gov/uk/courtdata/reporder/mapper/RepOrderMapperTest.java

JHunt16: Makes sense, I agree no need to duplicate as the mapping is already tested.

JHunt16 (approved):

PR #1049, opened 2024-09-11
LCAM-1545-get-rep-order

loganathanMOJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/Lcam-545)

## Checklist

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
@@ -102,10 +102,11 @@ public class RepOrderService {
 
     @Transactional
     public boolean exists(Integer repId, boolean hasSentenceOrderDate) {
-        log.info("Retrieve rep Order Count With Sentence Order Date - Transaction Processing - Start");
         if (!hasSentenceOrderDate) {
+            log.info("Retrieve rep Order count for repId: {}", repId);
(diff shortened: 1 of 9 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
```diff
@@ -101,8 +101,11 @@ public void update(Integer repId, Map<String, Object> repOrder) {
     }
 
     @Transactional
-    public boolean exists(Integer repId) {
+    public boolean exists(Integer repId, boolean hasSentenceOrderDate) {
         log.info("Retrieve rep Order Count With Sentence Order Date - Transaction Processing - Start");
```
nigelpain, line 105: Really minor thing, but wonder if it would be useful to print the `hasSentenceOrderDate` in the logs just to be clearer which query has been run?

nigelpain (commented):
Looks good, I do wonder if checking the rep order exists should be a separate endpoint as the function feels substantially different enough than getting the rep order details. However thats really out of the scope of this ticket 😃

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
loganathanMOJ: Thanks. Done now

nigelpain (approved):

PR #1050, opened 2024-09-11
LCAM-1414

muthusubra:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1414)

**NOTE: This PR was originally created to improve some of the DB code around persisting income evidence items. However, the original changes were subsequently reverted in the main branch as a result of the recent live issue that was causing income evidence items to be deleted unintentionally. Therefore, those earlier changes will need to be merged into this branch to resolve the conflicts shown below.**

- Simplify and improve logic around persisting income evidence items when updating a means assessment.
- New test cases to verify the persistence of income evidence items.
- Prevent updates to the applicant object when persisting an income evidence item.


maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/impl/FinancialAssessmentImpl.java
```diff
@@ -86,27 +86,29 @@ public FinancialAssessmentEntity update(FinancialAssessmentDTO financialAssessme
...
+        existingIncomeEvidences.forEach(evidence -> {
+            log.info("Existing evidence: {}", evidence.getIncomeEvidence());
+            // Check if the current evidence is not present in the updated financial assessment
+            if (financialAssessment.getFinAssIncomeEvidences().stream()
+                    .noneMatch(dto -> dto.getIncomeEvidence().equals(evidence.getIncomeEvidence()))) {
+                log.info("Deleting Financial Assessment Income Evidence Id: {}", evidence.getId());
+                finAssIncomeEvidenceRepository.deleteById(evidence.getId());
             }
-        }
+        });
```
Mhowell494, line 99: I think we can remove this code as:
`existingAssessment.getFinAssIncomeEvidences().clear()` should remove all existing records.

Mhowell494 (commented):
Not sure why SonarCloud hasn't flagged it, but we also need to enhance the existing unit and integration test scenarios.

maat-court-data-api/src/main/java/gov/uk/courtdata/assessment/impl/FinancialAssessmentImpl.java
muthusubra: Clear, will only clear the list and saveAndFlush won't delete the records. To simplify further added delete based on FIAS_ID to delete all the records as against deleting based on individual income evidence id

Mhowell494: AFAIK we just use `clear()` in the `HardshipReviewImpl` and it seems to work there?

muthusubra: Did it clear the delta(removed the records) from the table? I did test it and didn't seem to work. May be will try again

venkat980 (approved):

PR #1079, opened 2024-12-09
LCAM-1662|Enhance Find Rep_Order endpoint to return passport evidences

venkat980:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1662)

In order to be able to set the evidence received date correctly when updating the income evidence during the means assessment workflow, we need to check if the evidence has been added during a previous financial or passport assessment.

However we are currently unable to perform this check against passport evidence as this data is not returned in the RepOrderDTO from the find rep order endpoint in MAAT Court Data Api. Therefore, updated find RepOrder endpoint to return passport evidences.


maat-court-data-api/src/main/resources/db/changelog/changes/lcam_1662_grant_mla_crud_permissions_on_passport_ass_evidence.sql
```diff
@@ -0,0 +1 @@
+GRANT INSERT, SELECT, UPDATE, DELETE ON TOGDATA.PASSPORT_ASS_EVIDENCE TO MLA;
```
nigelpain, line 1: Do we just need SELECT privileges at the moment as we are only retrieving passport evidences for the find rep order endpoint?

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEvidenceEntity.java
```diff
@@ -0,0 +1,78 @@
...
+    @Column(name = "ID", nullable = false)
+    private Integer id;
+
+    @JsonBackReference
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "PAAS_ID", nullable = false)
+    private PassportAssessmentEntity passportAssessment;
+
+    @Column(name = "DATE_RECEIVED")
+    private LocalDateTime dateReceived;
```
nigelpain, line 38: I noticed on the table this and REMOVED_DATE are just DATE type, does this need to be defined as just LocalDate type in the code here?

maat-court-data-api/src/main/java/gov/uk/courtdata/dto/PassportAssessmentEvidenceDTO.java
```diff
@@ -0,0 +1,29 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+public class PassportAssessmentEvidenceDTO implements Serializable {
+    private Integer id;
+    private LocalDateTime dateReceived;
+    private LocalDateTime dateCreated;
+    private String userCreated;
+    private LocalDateTime dateModified;
+    private String userModified;
+    private String active;
```
nigelpain, line 22: The fields here such as `active`, `mandatory` and `adhoc` that are kept as Y/N Strings on the database, were we taking the approach of having them as Booleans in the DTOs and using a convertor to translate the values?

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEvidenceEntity.java
venkat980: That is correct, we are keeping this inline with the FIN_ASS_INCOME_EVIDENCE

maat-court-data-api/src/main/resources/db/changelog/changes/lcam_1662_grant_mla_crud_permissions_on_passport_ass_evidence.sql
nigelpain: Ok that makes sense, happy to resolve as we need to keep them both in line to make the logic in orchestrator simpler.

maat-court-data-api/src/main/java/gov/uk/courtdata/dto/PassportAssessmentEvidenceDTO.java
nigelpain: Ok that makes sense, happy to resolve as we need to keep them both in line to make the logic in orchestrator simpler.

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/PassportAssessmentEvidenceEntity.java
nigelpain: Ok that makes sense, happy to resolve as we need to keep them both in line to make the logic in orchestrator simpler.

nigelpain (approved):

PR #1088, opened 2025-01-29
DCES - Refactor Contributions Fetch to reduce DB impact.

LAA-Dave-Edwards:
## What

Refactoring the main dces contribution fetch to a split call to reduce db impact.
The Dev environment had a major issue with performance, and we noticed due to the exaggerated performance issues, that this specific call could be improved.

Doing the fetch top x with an id sort was magnitudes faster when other columns were excluded. So the call was split into 2 parts. The original call, which now only fetches the ids of those matching entries, and a use of the findByIdIn which will return the full details, but in a much more efficient manner.

## Checklist

## Additional checks

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/repository/ConcorContributionsRepository.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/repository/ConcorContributionsRepository.java
@@ -18,15 +18,16 @@ public interface ConcorContributionsRepository extends JpaRepository<ConcorContr
     List<ConcorContributionsEntity> findByIdIn(Set<Integer> ids);
 
     @Query("""
-       SELECT cc
+       SELECT cc.id
        FROM ConcorContributionsEntity cc
        WHERE cc.status = :status
        AND cc.id > :startId
        ORDER BY cc.id ASC""")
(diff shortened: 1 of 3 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/repository/ConcorContributionsRepository.java
```diff
@@ -18,16 +18,17 @@ public interface ConcorContributionsRepository extends JpaRepository<ConcorContr
     List<ConcorContributionsEntity> findByIdIn(Set<Integer> ids);
 
     @Query("""
-       SELECT cc
```
vvartakMoJ, line 21: Could a query like this have worked (and kept it to a single call)? (this is equivalent of what you showed me in the slack huddle this morning)

SELECT cc
FROM ConcorContributionsEntity cc
WHERE id in (
SELECT id FROM (
SELECT id
       FROM ConcorContributionsEntity cc
       WHERE cc.status = :status
       AND cc.id > :startId
       ORDER BY cc.id ASC
       )
       )

vvartakMoJ (commented):
Just a query to double-check my understanding.

PR #1091, opened 2025-02-04
DCES-694-Fdc-Fetch-Rework

LAA-Dave-Edwards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/DCES-694)

Refactoring the FDC get endpoints to use a native query.
Under current JPA method, if there are 200 Fdc Contributions found by the FindBy, when we map the object over to the response object. It will then perform 200 additional database queries to populate the RepOrderEntities.
While these are very simple queries ( select * from RepOrder where id = ? ) it both takes extra time, and hits the db unexpectedly.
The native query, and mapper will remove that additional burden, and keeps the response to what is needed.

## Checklist

## Additional checks

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
@@ -79,12 +79,12 @@ public class FdcContributionsService {
 
     public FdcContributionsResponse getFdcContributions(FdcContributionsStatus status) {
         log.info("Getting fdc contributions with status -> {}", status);
-        return buildFdcContributionsResponse(() -> fdcContributionsRepository.findByStatus(status));
+        return buildFdcContributionsResponse(() -> debtCollectionRepository.getFdcEntriesByStatus(status));
     }
 
     public FdcContributionsResponse getFdcContributions(List<Integer> fdcContributionIdList) {
         log.info("Getting fdc contributions for IDs in a list of size {}", fdcContributionIdList.size());
-        return buildFdcContributionsResponse(() -> fdcContributionsRepository.findByIdIn(new HashSet<>(fdcContributionIdList)));
+        return buildFdcContributionsResponse(() -> debtCollectionRepository.getFdcEntriesByIdIn(new HashSet<>(fdcContributionIdList)));
     }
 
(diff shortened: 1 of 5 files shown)


karlbaker02 (approved):
👍

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FdcContributionsEntity.java
```diff
@@ -96,7 +96,7 @@ public class FdcContributionsEntity {
     private String userModified;
 
     @JsonBackReference
-    @OneToOne(targetEntity = RepOrderEntity.class, fetch = FetchType.LAZY)
+    @OneToOne(targetEntity = RepOrderEntity.class, fetch = FetchType.EAGER)
```
dconneely, line 99: Did you mean to change this as well as change to JDBC template-based queries? If so, then looks fine to me.

dconneely (approved):
LGTM

maat-court-data-api/src/main/java/gov/uk/courtdata/entity/FdcContributionsEntity.java
LAA-Dave-Edwards: Yep, noticed that as well, reverted back to LAZY.

maat-court-data-api/src/main/java/gov/uk/courtdata/dces/service/FdcContributionsService.java
```diff
@@ -73,18 +73,18 @@ private FdcContributionEntry mapContributionEntry(FdcContributionsEntity entity)
                 .lgfsCost(entity.getLgfsCost())
                 .agfsCost(entity.getAgfsCost())
                 .maatId(entity.getRepOrderEntity().getId())
-                .sentenceOrderDate(Objects.nonNull(entity.getRepOrderEntity()) ? entity.getRepOrderEntity().getSentenceOrderDate() : null)
```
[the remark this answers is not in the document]
LAA-Dave-Edwards: As Fdcs always have RepOrders, it will always contain a MaatId.
Prior to this, it was being resolved via LAZY evaluation, which was the offending db calls this is removing.
Now the mapper is being used, which explicitly creates a RepOrderEntity, and attaches it to the FdcEntity as part of the process of resolving the DB call.

PR #1105, opened 2025-03-24
LCAM-1789: Fix Failing Evidence Fee Test Scenarios in the Functional Test Suite

Mhowell494:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1789)

Due to a recent Spring Boot upgrade, the framework now strictly adheres to HTTP standards by removing or overriding a manually set `Content-Length` headers when the response body is empty. This change broke our previous approach of using `Content-Length` headers to indicate the no. of records returned by some endpoints and therefore needed to be updated.

To resolve this the `CCOutcomeController` was refactored:
- Removed the dedicated `HEAD` endpoint, letting Spring Boot automatically handle `HEAD` requests via the `GET` endpoint.
- Updated the `GET` endpoint to include a custom `X-Total-Records` header that directly returns the record count without needing to serialise the full JSON response.

This solution avoids unnecessary overhead while providing clients with accurate metadata regarding the number of records.

- Several SonarQube issues were also addressed and formatting improved in several areas.

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
@@ -4,7 +4,6 @@ import gov.uk.courtdata.annotation.StandardApiResponse;
 import gov.uk.courtdata.dto.ErrorDTO;
 import gov.uk.courtdata.dto.RepOrderCCOutcomeDTO;
 import gov.uk.courtdata.enums.LoggingData;
-import gov.uk.courtdata.exception.ValidationException;
 import gov.uk.courtdata.model.RepOrderCCOutcome;
 import gov.uk.courtdata.reporder.service.CCOutcomeService;
 import gov.uk.courtdata.reporder.validator.CCOutComeValidationProcessor;
@@ -19,19 +18,17 @@ import java.util.List;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
-import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
-
-import static java.util.Objects.nonNull;
 
 
 @Slf4j
@@ -42,9 +39,7 @@ import static java.util.Objects.nonNull;
 public class CCOutcomeController {
 
     private final CCOutcomeService service;
-
     private final CCOutComeValidationProcessor validator;
-
 
     @PostMapping(produces = MediaType.APPLICATION_JSON_VALUE)
     @Operation(description = "Create a new RepOrder CC outcome")
@@ -62,14 +57,16 @@ public class CCOutcomeController {
                     schema = @Schema(implementation = ErrorDTO.class)
             )
     )
-    public ResponseEntity<RepOrderCCOutcomeDTO> create(@Parameter(description = "RepOrder CC outcome data",
-            content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
-                    schema = @Schema(implementation = RepOrderCCOutcome.class))) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
+    public ResponseEntity<RepOrderCCOutcomeDTO> create(
+            @Parameter(description = "RepOrder CC outcome data",
+                    content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
+                            schema = @Schema(implementation = RepOrderCCOutcome.class))) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
         LoggingData.MAAT_ID.putInMDC(repOrderCCOutCome.getRepId());
         log.info("Create Financial RepOrder CC outcome");
         validator.validate(repOrderCCOutCome);
         return ResponseEntity.ok(service.create(repOrderCCOutCome));
     }
+
 
     @PutMapping(produces = MediaType.APPLICATION_JSON_VALUE)
     @Operation(description = "Update a RepOrder CC outcome Record")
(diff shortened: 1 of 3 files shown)


maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
```diff
@@ -62,15 +57,17 @@ public class CCOutcomeController {
                     schema = @Schema(implementation = ErrorDTO.class)
             )
     )
-    public ResponseEntity<RepOrderCCOutcomeDTO> create(@Parameter(description = "RepOrder CC outcome data",
-            content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
-                    schema = @Schema(implementation = RepOrderCCOutcome.class))) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
+    public ResponseEntity<RepOrderCCOutcomeDTO> create(
+            @Parameter(description = "RepOrder CC outcome data",
+                    content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
+                            schema = @Schema(implementation = RepOrderCCOutcome.class))) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
```
karlbaker02, line 63: There is some weirdness going on here with the spacing by the looks of things, `description` `content` and `schema` should all be at the same indentation level I'd say.

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
```diff
@@ -87,17 +84,19 @@ public ResponseEntity<RepOrderCCOutcomeDTO> create(@Parameter(description = "Rep
...
     )
-    public ResponseEntity<RepOrderCCOutcomeDTO> update(@Parameter(description = "RepOrder CC outcome data",
-            content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
-                    schema = @Schema(implementation = RepOrderCCOutcome.class)
-            )
-    ) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
+    public ResponseEntity<RepOrderCCOutcomeDTO> update(
+            @Parameter(description = "RepOrder CC outcome data",
+                    content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
+                            schema = @Schema(implementation = RepOrderCCOutcome.class)
```
karlbaker02, line 90: Similar indentation comments as above

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
```diff
@@ -62,15 +57,17 @@ public class CCOutcomeController {
                     schema = @Schema(implementation = ErrorDTO.class)
             )
     )
-    public ResponseEntity<RepOrderCCOutcomeDTO> create(@Parameter(description = "RepOrder CC outcome data",
-            content = @Content(mediaType = MediaType.APPLICATION_JSON_VALUE,
-                    schema = @Schema(implementation = RepOrderCCOutcome.class))) @RequestBody RepOrderCCOutcome repOrderCCOutCome) {
+    public ResponseEntity<RepOrderCCOutcomeDTO> create(
```
karlbaker02, line 60: I don't see any behavioural changes in this commit - is that correct?

maat-court-data-api/src/test/java/gov/uk/courtdata/integration/repOrder/CCOutcomeControllerIntegrationTest.java
```diff
@@ -49,7 +49,7 @@ class CCOutcomeControllerIntegrationTest extends MockMvcIntegrationTest {
     @BeforeEach
     void setUp() {
         existingRepOrder = repos.repOrder.saveAndFlush(
-            TestEntityDataBuilder.getPopulatedRepOrder());
+                TestEntityDataBuilder.getPopulatedRepOrder());
```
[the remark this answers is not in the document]
karlbaker02: Ignore comment, I was going by the PR description and not commit description.

lkm0287:
I know it's kind of theoretical but it was bad that we were using content length for anything than its intended purpose to publish the message size in bytes.

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
Mhowell494: Not sure I'd agree because they're all parameters of the proceeding annotation, although I do prefer having the ending bracket on a new line.

karlbaker02: I do think there is something to be tidied up here, they are all parameters of the preceding annotation as you said but the indentation looks out to me. Not a major problem, just something to tidy up perhaps alongside everywhere else in the codebase that is in a similar state in a future PR.

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/controller/CCOutcomeController.java
karlbaker02: For my future self, same comments as previous around the `create` method in the CCOutcomeController.

karlbaker02 (commented):
I think there is some general refactoring that we can do at a later point (as it isn't brought in by this PR), around introducing consistency for GET methods to make them more RESTful. Some of the GET methods we have, such as for rep orders and CC outcomes, return the count as a header of the request, returning the actual entities in the response body. In other places, such as in the work queue offence controller, the GET methods return the count of entities directly in the response body only.

karlbaker02 (approved):
Looks good 👍 Some refactoring that can be done in the future, but this is outside the scope of this PR anyway.

Mhowell494:
> I think there is some general refactoring that we can do at a later point (as it isn't brought in by this PR), around introducing consistency for GET methods to make them more RESTful. Some of the GET methods we have, such as for rep orders and CC outcomes, return the count as a header of the request, returning the actual entities in the response body. In other places, such as in the work queue offence controller, the GET methods return the count of entities directly in the response body only.

Yes, agree - there isn't great consistency at the moment. In a perfect world, every endpoint should return data and then we'd be using `HEAD` requests to get a count of the records, but given that there were already several dedicated count methods, the only thing that really made sense at this stage was to switch over to `GET` in the meantime.

Mhowell494:
> One other thought: given we are changing existing endpoints which might be already in use by upstream services, there could potentially be a risk that these break when we deploy the Court Data API and before the relevant changes to those other services are also deployed? If this is a possibility and to be on the safe side should we find any issues with these Court Data API changes once deployed, it might be an idea to keep the old endpoints alongside the new ones until everything is deployed, at which point we can come back and clean up the old endpoints.

I think that's a reasonable suggestion, but it's probably easier to just have a period of downtime to coordinate all of the changes - I believe there is a MAAT release already planned for later this week.

karlbaker02:
> > One other thought: given we are changing existing endpoints which might be already in use by upstream services, there could potentially be a risk that these break when we deploy the Court Data API and before the relevant changes to those other services are also deployed? If this is a possibility and to be on the safe side should we find any issues with these Court Data API changes once deployed, it might be an idea to keep the old endpoints alongside the new ones until everything is deployed, at which point we can come back and clean up the old endpoints.
> 
> I think that's a reasonable suggestion, but it's probably easier to just have a period of downtime to coordinate all of the changes - I believe there is a MAAT release already planned for later this week.

Fair enough. Is this something we should have a release plan / general sequence of events for what to deploy when documenting on the Jira task just to make sure that we don't forget anything?

PR #1125, opened 2025-07-08
[LASB-4229] Add endpoint to delete all entries from maat_refs_to_extract

JHunt16:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-4229)

Added endpoint to delete all entries from maat_refs_to_extract table. 

Please note, I went with a DELETE sql command rather than TRUNCATE. I did this because it appears that only the table owner can truncate, or a user with the "drop any table" privilege; there is no way to grant a truncate privilege to a specific user like with select, delete etc. It is technically less-efficient to do it this way, but performance-wise it was negligible; I still managed to delete 267k records in half a second, and there shouldn't ever be more than that given this will run weekly.

## Checklist

## Additional checks

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/billing/controller/MaatReferenceExtractionController.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/billing/controller/MaatReferenceExtractionController.java
@@ -7,6 +7,7 @@ import io.swagger.v3.oas.annotations.tags.Tag;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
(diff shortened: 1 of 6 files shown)


maat-court-data-api/src/test/java/gov/uk/courtdata/integration/billing/MaatReferenceExtractionControllerIntegrationTest.java
```diff
@@ -36,12 +36,16 @@ void givenRecordsAlreadyExist_whenPopulateMaatReferencesToExtract_thenReturnErro
...
         assertTrue(
             runServerErrorScenario("The MAAT_REFS_TO_EXTRACT table already has entries",
-                getPostRequest()));
+                post(CREATE_ENDPOINT_URL)));
     }
 
-    private MockHttpServletRequestBuilder getPostRequest() {
-        return post(ENDPOINT_URL)
-            .contentType(MediaType.APPLICATION_JSON);
+    @Test
```
karlbaker02, line 42: Maybe we could include another test that returns a `400` response if there's some exception somewhere down the stack (mimicking that the transaction failed)?

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/controller/MaatReferenceExtractionController.java
```diff
@@ -29,4 +30,14 @@ public ResponseEntity<Object> populateMaatReferencesToExtract() {
         log.info("Populate MAAT references Request completed successfully");
         return ResponseEntity.ok().build();
     }
+
+    @DeleteMapping(value = "/delete-maat-references")
```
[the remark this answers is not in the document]
JHunt16: Makes sense - don't know why I did it that way!

maat-court-data-api/src/test/java/gov/uk/courtdata/integration/billing/MaatReferenceExtractionControllerIntegrationTest.java
JHunt16: As discussed, I'm not sure how we can do that in the integration test as it has no mocks and my delete method on the service has no explicit conditions that throw exceptions - I guess the db connection could drop or something, but I'm not sure how to mimic that in this particular test class.

I have added an additional test to the service though, just ensuring if an exception occurs in the repo the service throws it - might be somewhat helpful for a bit of extra assurance?

PR #1128, opened 2025-07-10
LASB-4224 Create new endpoint to reset send_to_cclf flag in applicants table

tharunmoj:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-4224)

Created new endpoint to reset send_to_cclf flag in applicants table.

## Checklist

## Additional checks

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/billing/entity/BillingApplicantEntity.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/billing/entity/BillingApplicantEntity.java
@@ -59,8 +59,4 @@ public class BillingApplicantEntity {
 
     @Column(name = "user_modified")
     private String userModified;
-
-    @Column(name = "SEND_TO_CCLF")
-    private String sendToCclf;
-
 }


maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepository.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Collections;
+
+@Repository
+@RequiredArgsConstructor
+public class ApplicantCclfResetRepository {
+
+    private static final String RESET_CCLF_SQL = """
+        UPDATE APPLICANTS
```
karlbaker02, line 15: You'll probably run into the same problem I did here, in that you'll need to prefix the table with the `TOGDATA` schema to not run into issues.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepository.java
```diff
@@ -0,0 +1,28 @@
...
+        SET SEND_TO_CCLF = NULL,
+            DATE_MODIFIED = SYSDATE,
+            USER_MODIFIED = USER
+        WHERE ID IN (SELECT APPL_ID FROM MAAT_REFS_TO_EXTRACT)
```
karlbaker02, line 19: Again up for debate, but the way we've done this for applicant history and rep orders is to pass in a list of ids as part of the request, rather than selecting from the table. We actually need to go with this approach too for applicant and applicant history records, for the reasons detailed below:

- when we extract applicant records, they're extracted based on a join to the `MAAT_REFS_TO_EXTRACT` table _as well as_ a union against everything in the `APPLICANTS` table where `send_to_cclf`=`Y`. The above will only reset the first half and it's highly likely that the data in the `APPLICANTS` table will have changed since the extract was run that it would now reset additional records which weren't part of the extract. For that reason the ids to reset need to be passed in.
- we also can't have confidence that the consumer of this service isn't going to encounter some issue of it's own as part of calling MAAT API to reset these records, so to ensure that we don't have any partial updates that might not be expected by the consumer, we should again pass in the ids to reset.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepository.java
```diff
@@ -0,0 +1,28 @@
...
+        """;
+
+    private final NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Transactional
+    public void resetApplicantCclfFlag() {
+        jdbcTemplate.update(RESET_CCLF_SQL, Collections.emptyMap());
```
karlbaker02, line 26: For the update statement as it currently stands, I don't think it's necessary to use the `NamedParameterJdbcTemplate`, especially as we're passing it zero parameters. Potentially for the revised update statement as per my comments above, but it might also be best to stick with how we are doing this sort of update in a few other places throughout the codebase without the use of this template.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/service/ApplicantCclfResetService.java
```diff
@@ -0,0 +1,16 @@
+package gov.uk.courtdata.billing.service;
+
+import gov.uk.courtdata.billing.repository.ApplicantCclfResetRepository;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
```
karlbaker02, line 7: I think you might also be missing the `@Slf4j` annotation here

maat-court-data-api/src/test/java/gov/uk/courtdata/billing/controller/ApplicantCclfResetControllerTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.verify;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@WebMvcTest(ApplicantCclfResetController.class)
+@AutoConfigureMockMvc(addFilters = false)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+class ApplicantCclfResetControllerTest {
```
karlbaker02, line 21: These tests will need to change once we change to pass in a list of ids, so no comments here until then

maat-court-data-api/src/test/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepositoryTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.Mockito.*;
+
+class ApplicantCclfResetRepositoryTest {
+
+    @Test
+    void shouldExecuteStoredProcedureSuccessfully() {
+        NamedParameterJdbcTemplate mockJdbcTemplate = Mockito.mock(NamedParameterJdbcTemplate.class);
```
karlbaker02, line 15: Similar comment as to previous that I don't think we need to introduce `NamedParameterJdbcTemplate`.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepository.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Repository
+@RequiredArgsConstructor
+public class ApplicantCclfResetRepository {
+
+    private static final String RESET_CCLF_SQL = """
+        UPDATE APPLICANTS
+        SET SEND_TO_CCLF = NULL,
+            DATE_MODIFIED = SYSDATE,
+            USER_MODIFIED = USER
```
karlbaker02, line 18: For the other endpoints we have gone with passing in the username. The ask here hasn't been particularly clear but it feels like it makes sense - suggest maybe going with that as well for now, if we decide to change in the future then all the endpoints can be changed at once.

karlbaker02 (changes requested):
Thanks for this Tharun! The main thing that needs changing here is to pass in a list of ids of the applicant records that need to be reset as we're not currently catering for the union part of the data extract as things stand.

You might also consider renaming the controller, service, repository to be aligned with those we've already created (see latest version of the `main` branch in the `billing` package), which means that we can build on these classes in the future if we need to do anything else in the billing space with these resources - as they stand they're all closely coupled to CCLF-specific actions.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/ApplicantCclfResetRepository.java
tharunmoj: @karlbaker02 Can you please update the ticket, I didn't aware of this requirement.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/controller/ApplicantResetController.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.bind.annotation.*;
+
+@Slf4j
+@RestController
+@RequiredArgsConstructor
+@Tag(name = "Applicant CCLF Reset", description = "Rest API for resetting SEND_TO_CCLF flag in applicants")
+@RequestMapping("${api-endpoints.billing-domain}/applicant")
+public class ApplicantResetController {
+
+    private final ApplicantResetService applicantResetService;
```
karlbaker02, line 21: I was going to ask if you could rename all of these classes that start `ApplicantReset` to be something more like `ApplicantBilling ...`, but then I realised that we do _already_ have a controller in the code for dealing with applicant billing requests. You may have missed it as it's naming is not consistent with the others - it's called `BillingApplicantController`; this is what we should be using for the applicant logic, with the _reset_ being one endpoint on that controller.

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/controller/ApplicantResetController.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @PatchMapping("/reset")
```
karlbaker02, line 23: If you can change the `BillingApplicantController`'s `@RequestMapping` on line 21 to be `@RequestMapping("${api-endpoints.billing-domain}/applicants")` then I think you can get rid of the `"/reset"` mapping here (you'll also need to remove `"/applicants"` on line 26 of the controller for existing tests to pass).

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/request/ApplicantResetRequest.java
```diff
@@ -0,0 +1,17 @@
...
+
+@Data
+public class ApplicantResetRequest {
+
+    @NotBlank(message = "Username must not be blank")
+    private String username;
+
+    @NotEmpty(message = "Applicant IDs must not be empty")
+    private List<Integer> applicantIds;
+}
```
karlbaker02, line 17: I would make use of the existing `UpdateRepOrderBillingRequest` class for now and remove this one, as in Nigel's [pending PR](https://github.com/ministryofjustice/laa-maat-court-data-api/pull/1126) this class has been refactored to make a generic `UpdateBillingRequest` that can be used across different resources).

maat-court-data-api/src/test/java/gov/uk/courtdata/billing/controller/ApplicantResetControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Autowired
+    private MockMvc mvc;
+
+    @MockitoBean
+    private ApplicantResetService applicantResetService;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Test
```
karlbaker02, line 33: It might be worth adding another test to check for invalid input (400 response returned).

maat-court-data-api/src/test/java/gov/uk/courtdata/billing/repository/ApplicantResetRepositoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+            ID INT PRIMARY KEY,
+            SEND_TO_CCLF VARCHAR(1),
+            DATE_CREATED TIMESTAMP,
+            USER_CREATED VARCHAR(100),
+            DATE_MODIFIED TIMESTAMP,
+            USER_MODIFIED VARCHAR(100)
+        );
+        """
+})
+class ApplicantResetRepositoryTest {
```
karlbaker02, line 30: Very cool, I wasn't aware we had the capacity to test at the repository level. Nice 💯

maat-court-data-api/src/main/java/gov/uk/courtdata/billing/entity/BillingApplicantEntity.java
```diff
@@ -60,4 +60,7 @@ public class BillingApplicantEntity {
     @Column(name = "user_modified")
     private String userModified;
 
+    @Column(name = "SEND_TO_CCLF")
```
lkm0287, line 63: Can you please not amend this entity - this will break my endpoint as this entity is tied to the native query in https://github.com/ministryofjustice/laa-maat-court-data-api/blob/89febdfec8403edbf1924182f991315e539fa738/maat-court-data-api/src/main/java/gov/uk/courtdata/billing/repository/BillingApplicantRepository.java

tharunmoj: Thanks @lkm0287 ,

I need the SEND_TO_CCLF flag for the logic. So, Shall I add SEND_TO_CCLF in your select statement or I need to declare a new entity and repository. Please advice.

tharunmoj: Moved the logic. Its used in BillingApplicantUpdateRepository and BillingApplicantUpdateRepositoryTest.

karlbaker02 (approved):
Looks good, thanks Tharun! 👍

PR #1163, opened 2026-04-13
LCAM-1995 - Adding Initial v2 Create Passported Endpoint.

LAA-Dave-Edwards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LCAM-1995)

Adding a new v2 endpoint for the create passported assessment, and associated mapper and tests.
Updated ProblemDetailExceptionHandler to ensure the v2 controller is also handled, and updated tests to ensure ProblemDetail responses.
Mapper still has outstanding question w.r.t. under18 court type. Awaiting further information. 
Tests have been added to cover the new areas, and ensure mapping is behaving correctly. 

## Checklist

## Additional checks


Mhowell494 (changes requested):

maat-court-data-api/src/test/java/gov/uk/courtdata/passport/controller/PassportAssessmentControllerV2Test.java
```diff
@@ -44,7 +49,45 @@ void givenNonExistentLegacyPassportAssessmentId_whenFindIsCalled_thenReturnBadRe
...
+    @Test
+    void givenAPartialRequest_whenCreateIsCalled_thenReturnValidationFailureResponse() throws Exception {
+        var request = TestModelDataBuilder.buildValidPopulatedCreatePassportedAssessmentRequest(false);
+        request.getPassportedAssessment().setAssessmentReason(null);
+        var expectedEntity = TestModelDataBuilder.buildValidCreatePassportedAssessmentResponse();
+        when(passportAssessmentService.create(any())).thenReturn(expectedEntity);
+        mvc.perform(MockMvcRequestBuilders.post(ENDPOINT_URL)
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .content(objectMapper.writeValueAsString(request)))
+                .andExpect(status().isBadRequest())
```
Mhowell494: It would be better if we also checked at least one field in the problem response, and ideally verified the service is not called if bean validation is supposed to fail before the service layer. You can probably copy this from the IOJ appeal controller tests.

maat-court-data-api/src/test/java/gov/uk/courtdata/passport/service/PassportAssessmentServiceV2Test.java
```diff
@@ -36,4 +62,97 @@ void whenFindByLegacyIdIsInvoked_thenPassportedAssessmentIsRetrieved() {
...
+        when(passportAssessmentPersistenceService.save(entity)).thenReturn(entity);
+        when(applicantService.find(APPLICANT_ID)).thenReturn(partner);
+        when(passportAssessmentMapper.toApiCreatePassportedAssessmentResponse(entity)).thenReturn(response);
+
+        passportAssessmentService.create(request);
+
+        verify(repOrderService).exists(REP_ID);
+        verify(passportAssessmentMapper).toPassportAssessmentEntity(request);
+        verify(passportAssessmentPersistenceService).save(entity);
+        verify(assessmentReplacementService).replacePreviousAssessments(any(PassportAssessmentEntity.class));
```
Mhowell494: We have the entity available so this could be stricter:

```java
verify(assessmentReplacementService).replacePreviousAssessments(entity);
```

Mhowell494 (changes requested):

Mhowell494 (changes requested):
Looking good overall, just a few minor tweaks to tidy things up. I think a couple of my earlier comments in the thread may have been missed as well.

maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/service/RepOrderService.java
```diff
@@ -61,11 +62,17 @@ public RepOrderDTO find(Integer repId, boolean hasSentenceOrderDate) {
...
+        return repOrderMapper.repOrderEntityToRepOrderDTO(updateDateCompleted(updateAppDateCompleted.getRepId(), updateAppDateCompleted.getAssessmentDateCompleted()));
+    }
+
+    @Transactional
+    public RepOrderEntity updateDateCompleted(Integer repId, LocalDateTime dateCompleted) {
         log.info("update app date completed - Transaction Processing - Start");
-        return repOrderMapper.repOrderEntityToRepOrderDTO(repOrderImpl
-                .updateAppDateCompleted(updateAppDateCompleted.getRepId(), updateAppDateCompleted.getAssessmentDateCompleted()));
+        return repOrderImpl
+                .updateAppDateCompleted(repId, dateCompleted);
```
Mhowell494: I’d avoid doing this as part of persisting a passported assessment. The legacy stored procedure applies additional conditions around case type and full assessment status that aren’t represented here.

This decision is better owned by the orchestrator. It can determine whether the application completion date should be set and apply it consistently. As I understand it, Nigel is planning to handle this via the RepOrder patch endpoint as part of his changes.

Mhowell494 (changes requested):

Mhowell494 (approved):

PR #1168, opened 2026-04-27
LASB-4996: Rep Order search enhancements

adamwlb-moj:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-4996)

This PR contains two improvements to the MAAT rep order search API, at the request of the ACD team.

- Makes date of birth an optional search parameter
   - Updates `dob` field on `MaatSearchRequest` so it can be null.
   - Updates database query to accomodate no DOB being passed.
   - Adds new `@DataJpaTest` for this repository to verify this change.
   - Adds unit tests for this scenario.
- Retrieves case URN from common platform
   - Populates `caseUrn` in `LinkingDetail` response from an additional call to common platform (`RepOrderCPDataRepository`), searching with the MAAT ID. (Rather than the link table).
   - Add unit tests for this scenario.
   - Updates test helper builder methods to facilitate the new tests.
   - Updates `RepOrderControllerIntegrationTest` to reflect the new flow.

## Checklist

- [X] Tests should be passing: `./gradlew test`
- [X] Github should not be reporting conflicts; you should have recently run `git rebase main`.
- [X] Avoid mixing whitespace changes with code changes in the same commit. These make diffs harder to read and conflicts more likely.
- [X] You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- [X] You should have checked that the commit messages say why the change was made.

## Additional checks

### DIFF
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/model/reporder/MaatSearchRequest.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/model/reporder/MaatSearchRequest.java
@@ -19,7 +19,6 @@ public class MaatSearchRequest {
     private String firstName;
     @NotNull
     private String lastName;
-    @NotNull
     private LocalDate dob;
     private String niNumber;
     @NotNull
--- a/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/mapper/RepOrderMapper.java
+++ b/maat-court-data-api/src/main/java/gov/uk/courtdata/reporder/mapper/RepOrderMapper.java
@@ -83,7 +83,7 @@ public interface RepOrderMapper {
                 .build();
     }
 
-    default MaatSearchResponse mapMaatSearchResponse(Integer maatId, List<WqLinkRegisterEntity> wqList) {
+    default MaatSearchResponse mapMaatSearchResponse(Integer maatId, List<WqLinkRegisterEntity> wqList, String caseUrn) {
         if (wqList == null || wqList.isEmpty()) {
             return MaatSearchResponse.builder()
                     .maatId(maatId)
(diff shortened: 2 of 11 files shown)


loganathanMOJ:
We have an existing integration test, givenAValidInput_whenSearchApplicationIsInvoked_thenCorrectResponseIsReturned, in RepOrderControllerIntegrationTest. Currently, the case URN is being retrieved from the linking table, but this needs to be changed to retrieve it from the Common Platform table And RepOrderRepositoryTest might not required

adamwlb-moj:
> We have an existing integration test, givenAValidInput_whenSearchApplicationIsInvoked_thenCorrectResponseIsReturned, in RepOrderControllerIntegrationTest. Currently, the case URN is being retrieved from the linking table, but this needs to be changed to retrieve it from the Common Platform table And RepOrderRepositoryTest might not required

That existing test was failing without my update to the integration test to add a record to the common platform table
```
repos.repOrderCPData.save(TestEntityDataBuilder.getRepOrderEntity(REP_ORDER_ID_NO_SENTENCE_ORDER_DATE));
```

Since I've changed the logic, none of the URNs are retrieved from the linking table, they all come from the common platform table now 🙂 

On `RepOrderRepositoryTest` - I think it's useful to be able to verify this behaviour as a unit test? Relying on the integration test seems very brittle to me. When I changed the SQL query nothing broke in the existing tests, but a null DOB would have caused a null pointer.

loganathanMOJ (approved):
We have an existing integration test, givenAValidInput_whenSearchApplicationIsInvoked_thenCorrectResponseIsReturned, in RepOrderControllerIntegrationTest. Currently, the case URN is being retrieved from the linking table, but this needs to be changed to retrieve it from the Common Platform table and  the new test RepOrderRepositoryTest might not required

aroberts10 (approved):

flynnbops (approved):
Nice work on test coverage with the change

